PR #47, opened 2019-04-24
Don't catch every exception.

### LINKED ISSUE #46: When the server can't connect to MySQL, it tells the user that their password is wrong
This is a bad user experience, because it seems like their password was incorrect, but the log shows the real error.

Instead it should throw a real exception.

The exception that should be thrown is BadGateway.

Steps to reproduce:

1. Start server
2. Turn off mysql
3. Try logging in
4. Look at the log.

The reason the real mysql exception gets turned into a "Incorrect Username or Password" is because _all_ exceptions are being caught in `/src/login/controller.ts`

juhangsin:
Fixes #46

### DIFF
--- a/src/login/controller.ts
+++ b/src/login/controller.ts
@@ -23,7 +23,6 @@ class LoginController extends Controller {
     try {
       user = await UserService.findByIdentity('mailto:' + ctx.request.body.username);
     } catch (err) {
-      console.log('hello world', err)
       if (err instanceof NotFound) {
         log(EventType.loginFailed, ctx);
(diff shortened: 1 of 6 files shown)


src/login/controller.ts
```diff
@@ -22,8 +23,13 @@ class LoginController extends Controller {
     try {
       user = await UserService.findByIdentity('mailto:' + ctx.request.body.username);
     } catch (err) {
-      log(EventType.loginFailed, ctx);
-      return this.redirectToLogin(ctx, 'Incorrect username or password');
+      console.log('hello world', err)
```
evert: Debug code -_-

PR #48, opened 2019-04-25
Registration form

### LINKED ISSUE #32: Add a HTML registration form
Allow new users to register.

First steps:

1. Add a link from the 'home document' to `/register`
2. Add an exception to the authentication middleware to allow access to `/register` without login
2. Create a new route for `/register`
3. Create a new controller : `/src/register/controller.ts`
4. Create a method in the controller `get()`
5. Render a HTML form

### LINKED ISSUE #33: Allow registration to be disabled
This should be done after #32 

We basically want a environment variable flag that allows registration to be turned on or off.

juhangsin:
* Fixes #32 
* Fixes #33


evert:
Still needs to create the actual new user!

PR #52, opened 2019-07-05
Feature, Resource-based privileges

mihok:
### Changes

- Adding a resource column in the `user_privileges` table
- Updating the `privilege` column to `scope` for more clarity since both the scope and resource together make up a privilege now
- Updated some functions to leverage resource in addition to scope
- `getPrivilegesForUser` now returns a map of resource to array of scopes

### Breaking Change

So for this change, the format of the returned privileges object has changed. Before it was a simple array of scopes available to the user that is authorized:
```
['foo', 'bar', 'bar-write']
```

And now it is a map of resources and scopes:
```
{
  '*': ['foo', 'bar'],
  'http://localhost:8000/my/resource': ['bar-write'],
}
```

Where `*` represents any resource. I opted to prefer `*` over `NULL` for both the table column value as well as in the returned map of resources because it ended up unnecessarily complicating things by trying to have `NULL` represent any resource in the data structure, but then convert it to some other identifier in the returned map of privileges.

### DIFF
--- a/src/privilege/service.ts
+++ b/src/privilege/service.ts
@@ -1,12 +1,13 @@
 import db from '../database';
 import { User } from '../user/types';
+import { PrivilegeMap } from './types';
 
 type PrivilegeRow = {
   resource: string,
   scope: string,
 };
 
-export async function getPrivilegesForUser(user: User): Promise<any> {
+export async function getPrivilegesForUser(user: User): Promise<PrivilegeMap> {
 
   const query = 'SELECT resource, scope FROM user_privileges WHERE user_id = ?';
   const result = await db.query(query, [user.id]);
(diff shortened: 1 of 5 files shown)


src/privilege/service.ts
```diff
@@ -2,26 +2,37 @@ import db from '../database';
 import { User } from '../user/types';
 
 type PrivilegeRow = {
-  privilege: string;
+  resource: string,
+  scope: string,
 };
 
-export async function getPrivilegesForUser(user: User): Promise<string[]> {
+export async function getPrivilegesForUser(user: User): Promise<any> {
```
evert: `any` should really be typed.

mihok:
Not sure why the Travis CI Branch test hangs?

evert:
idk sometimes it happens!

PR #60, opened 2019-09-19
added confirming password for user signup

juhangsin:


src/register/controller.ts
```diff
@@ -16,6 +16,15 @@ class RegistrationController extends Controller {
 
   async post(ctx: Context) {
 
+    const userPassword = ctx.request.body.password;
```
evert: Given that you're putting the password in its own variable here, then line 44 should also use it

PR #74, opened 2019-10-04
Camel case

### LINKED ISSUE #72: Change templates to camel-case

juhangsin:
Changed to camel case. please take a look at the token.ts one since I'm not 100% sure if they were associated.

Fixes #72

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,18 @@
     "assets",
     "templates"
   ],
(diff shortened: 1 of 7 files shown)


evert:
You were right, the token controller can't be changed. The reason is that this is part of the OAuth2 protocol and we have to match the specification exactly to not break anything

PR #78, opened 2019-10-08
Reset Password, created new routes & template. No post call included yet

juhangsin:
Comepleted
When the user clicks Forgot Password btn, it direct you to a forgot password page (O)
From the forgot password page, it will ask to type your email address (O)

Not Completed Yet
User clicks the submit button, there will be a notice that tells you email has been sent, and the user receives an email with URL (X)
User clicks the url, and re-directs you to the Create Password page. (X)

### DIFF
--- a/templates/login.hbs
+++ b/templates/login.hbs
@@ -19,6 +19,10 @@
     <input type="hidden" name="{{@key}}" value="{{this}}" />
   {{/each}}
 
+  <p class="reset-password">
+    <a href="/reset-password">Reset Password</a>
+  </p>
+
   {{#if registrationEnabled}}
   <p class="login-options">
     <a href="/register">Register</a>
--- a/templates/login.hbs
+++ b/templates/login.hbs
@@ -18,14 +18,11 @@
   {{#each hiddenFields}}
     <input type="hidden" name="{{@key}}" value="{{this}}" />
   {{/each}}
-
-  <p class="reset-password">
-    <a href="/reset-password">Reset Password</a>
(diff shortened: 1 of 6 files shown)


templates/login.hbs
```diff
@@ -19,6 +19,10 @@
     <input type="hidden" name="{{@key}}" value="{{this}}" />
   {{/each}}
 
+  <p class="reset-password">
+    <a href="/reset-password">Reset Password</a>
```
evert: Lets make the label 'Forgot password'.

Also, this setting should be inside of the `login-options` `<p>` tag, so they are on the same line with the same style.

templates/reset-password.hbs
```diff
@@ -0,0 +1,16 @@
+{{# if msg}}
+<div class="msg">{{ msg }}</div>
+{{/if}}
+
+<form action="{{ action }}" method="post">
+
+  <label for="emailaddress">Email</label>
```
evert: camelCase

templates/reset-password.hbs
```diff
@@ -0,0 +1,16 @@
+{{# if msg}}
+<div class="msg">{{ msg }}</div>
+{{/if}}
+
```
evert: It would be nice if there was a better description on this page. For example:

```
<p>You forgot your password! Don't worry, it happens to the best of us.</p>
<p>To reset your password, enter your email address below. If we had a record of your email address, we will send you an email with a link that allows you to reset your password.</p>
```

PR #79, opened 2019-10-09
Send email for reset password function using nodemailer. Haven't actually got the email yet, but the log says it should be successfully sending the email

juhangsin:


src/reset-password/controller.ts
```diff
@@ -0,0 +1,38 @@
...
+            ctx.response.headers.set('location', '/reset-password?msg=We+can\'t+seem+to+find+your+record.+Please+try+gain');
+            return;
+        } else {
+            throw err;
+        }
+    }
+
+    await sendResetPasswordEmail(user);
+    ctx.status = 303;
+    ctx.response.headers.set('location', '/reset-password?msg=We\'ve+sent+you+a+link+to+your+email+for+changing+password');
```
evert: For a future PR it would be nice to have better confirm pages, but not needed right now

src/reset-password/service.ts
```diff
@@ -0,0 +1,20 @@
...
+
+
+export async function sendResetPasswordEmail(user: User) {
+
+    // create reusable transporter object using the default SMTP transport
+    const transporter = nodemailer.createTransport(process.env.SMTP_URL);
+
+    // send mail with defined transport object
+    const info = await transporter.sendMail({
+        from: '"John Doe" <system@pegasi.badgateway.net>', // sender address
```
evert: This should be an environment variable as well.

src/reset-password/service.ts
```diff
@@ -0,0 +1,20 @@
...
+        to: user.identity.substring(7), // list of receivers
+        subject: 'Reset Password Request', // Subject line
+        text: 'Please click below link to reset your password!', // plain text body
```
evert: This can be done in a future PR, but the emails might need to be improved a bit =)

src/reset-password/service.ts
```diff
@@ -0,0 +1,20 @@
+import nodemailer from 'nodemailer';
+import { User } from '../user/types';
+
+
+export async function sendResetPasswordEmail(user: User) {
+
+    // create reusable transporter object using the default SMTP transport
+    const transporter = nodemailer.createTransport(process.env.SMTP_URL);
```
evert: Throw an error if SMTP_URL is not set...

src/reset-password/service.ts
evert: Maybe SMTP_EMAIL_FROM

src/reset-password/service.ts
evert: Or SMTP_EMAIL_FROM

src/reset-password/service.ts
```diff
@@ -0,0 +1,27 @@
+import nodemailer from 'nodemailer';
+import { User } from '../user/types';
+
+export async function sendResetPasswordEmail(user: User) {
+
+    if (!process.env.SMTP_EMAIL_FROM) {
+        throw new Error('The environment variable SMTP_EMAIL_FROM must be set');
+    }
+
+    if (!process.env.SMPT_URL) {
```
evert: `SMTP`

src/reset-password/service.ts
```diff
@@ -0,0 +1,27 @@
...
+        throw new Error('The environment variable SMTP_URL must be set. Needs to contain "smtps://[Username]:[Password]@[Host]:[Port]"');
+    }
+
+    const transporter = nodemailer.createTransport(process.env.SMTP_URL);
+
+
+    // send mail with defined transport object
+    const info = await transporter.sendMail({
+        from: '"Admin" <' + process.env.SMTP_EMAIL_FROM + '>', // sender address
```
evert: I think it would be even nicer if the entire 'from' string can be taken from SMTP_EMAIL FROM, not just the email address

PR #82, opened 2019-10-18
Added new sql stat & created new func for token

juhangsin:


src/reset-password/service.ts
```diff
@@ -25,3 +28,28 @@ export async function sendResetPasswordEmail(user: User) {
...
+    const token = crypto.randomBytes(32).toString('base64url').replace('=', '');
+    const query = 'INSERT INTO reset_password_token SET user_id = ?, token = ?, created_at = UNIX_TIMESTAMP()';
+    await database.query(query, [
+        user.id,
+        await bcrypt.hash(token, 12),
+    ]);
+    return token;
+}
+
+export async function validateToken(user: User, token: string): Promise<boolean> {
```
evert: This function should prevent a token from being uses that older than a set amount of time. For example, 2 hours. It would be nice if a specific error would be thrown in this case.

src/reset-password/service.ts
```diff
@@ -25,3 +28,28 @@ export async function sendResetPasswordEmail(user: User) {
...
+    const query = 'SELECT token FROM reset_password_token WHERE user_id = ?';
+    const result = await database.query(query, [user.id]);
+
+    const hashes: string[] = result[0].map( (row: { token: string }) => row.token );
+
+    for (const hash of hashes) {
+        if (await bcrypt.compare(token, hash)) {
+            return true;
+        }
+    }
```
nchangfong: after validating a token, is there any mechanism by which to mark it as validated? i.e., that the account recovery is completed?

evert: this should be added

nchangfong: a suggestion: change the `created_at = UNIX_TIMESTAMP()` field to `expires_at = UNIX_TIMESTAMP() + TTL` and then, after validation, set `expires_at = UNIX_TIMESTAMP()`

evert: Agree with adding a new field, but I think we can just delete the record after it was used?

nchangfong: is there another mechanism to audit one's account and see the last time a password was reset or a reset was requested?

evert: Yes, we have audit logs and we should add these as new events! good idea. Probably need 2 types of events:

* password reset requested
* password reset completed

mysql-schema/026-reset-password-token.sql
```diff
@@ -7,7 +7,8 @@ CREATE TABLE reset_password_token (
   id INT UNSIGNED NOT NULL PRIMARY KEY AUTO_INCREMENT,
   user_id INT UNSIGNED NOT NULL,
   token VARCHAR(50) NOT NULL UNIQUE,
-  created_at INT UNSIGNED NOT NULL
+  expires_at INT UNSIGNED NOT NULL,
+  created_at INT UNSIGNED NOT NULL,
```
evert: This is gonna error

src/reset-password/controller.ts
```diff
@@ -30,9 +32,13 @@ class ResetPasswordController extends Controller {
...
+    await log(EventType.resetPasswordRequest, ctx.ip(), user.id);
+
     ctx.status = 303;
     ctx.response.headers.set('location', '/reset-password?msg=We\'ve+sent+you+a+link+to+your+email+for+changing+password');
   }
 }
 
 export default new ResetPasswordController();
+
+// log(EventType.resetPasswordSuccess, ctx);
```
evert: Debug code

PR #86, opened 2019-11-06
Password reset email template with a provided URL

juhangsin:


src/reset-password/service.ts
```diff
@@ -17,22 +18,28 @@ export async function sendResetPasswordEmail(user: User) {
     }
 
     const transporter = nodemailer.createTransport(process.env.SMTP_URL);
-
+    const token = await createToken(user);
+    const emailTemplate =
+    render('emails/reset-password-email', {
```
evert: The indentation is weird for this. I think I would put `render` on the line above.

src/reset-password/service.ts
```diff
@@ -17,22 +18,28 @@ export async function sendResetPasswordEmail(user: User) {
...
+        name: user.nickname,
+        url: 'https://auth-server.example/reset-password/token/' + token,
```
evert: Url is definitely wrong right now... We can fix this in a future PR... not 100% sure how to do this yet.

PR #87, opened 2019-11-07
Better error message (#70)

juhangsin:


src/register/controller.ts
```diff
@@ -26,8 +26,11 @@ class RegistrationController extends Controller {
     }
 
     try {
-      await userService.findByIdentity('mailto:' + ctx.request.body.emailAddress);
-      throw new Error('User already exists');
+      if (await userService.findByIdentity('mailto:' + ctx.request.body.emailAddress)) {
+        ctx.status = 303;
+        ctx.response.headers.set('Location', '/register?msg=User+already+exists.+Please+use+a+different+email');
```
evert: How about something like... 'User already exists, did you forget your password?'

juhangsin: And maybe in the future, we can have a link that goes to the forgot password page?

PR #89, opened 2019-11-07
After validating token, redirects to change pw

juhangsin:


src/reset-password/controller/redirect.ts
```diff
@@ -0,0 +1,8 @@
+import Controller from '@curveball/controller';
+
+class ChangePasswordController extends Controller {
+
+
+}
+
+export default new ChangePasswordController();
```
juhangsin: Temporary file

src/app.ts
```diff
@@ -30,4 +30,8 @@ import { load } from './server-settings';
 
   console.log('Listening on port', port);
 
+  if (!process.env.PUBLIC_URI) {
+    process.env.PUBLIC_URI = 'http://localhost:' + port + '/';
```
evert: These new lines should go above `app.listen`, and I  think it should also have a console.log, like `PUBLIC_URI environment variable was not set, defaulting to x`

src/reset-password/service.ts
```diff
@@ -43,24 +43,21 @@ export async function createToken(user: User): Promise<string> {
...
+        token,
         tokenTTL
     ]);
     return token;
 }
 
-export async function validateToken(user: User, token: string): Promise<boolean> {
-    const query = 'SELECT token FROM reset_password_token WHERE user_id = ? AND token = ? AND expires_at > UNIX_TIMESTAMP()';
-    const result = await database.query(query, [user.id]);
+export async function validateToken(token: string): Promise<User> {
```
evert: One of the requirements for this function is that tokens can only be used once. So when a token is 'valid', it should be deleted from the table.

src/reset-password/service.ts
```diff
@@ -43,24 +43,21 @@ export async function createToken(user: User): Promise<string> {
...
+    const result = await database.query(query, [token]);
 
-    const hashes: string[] = result[0].map( (row: { token: string }) => row.token );
-
-    for (const hash of hashes) {
-        if (await bcrypt.compare(token, hash)) {
-            return true;
-        }
+    if (result[0].length !== 1) {
+      throw new BadRequest ('Invaild Token');
```
evert: Needs a better message

src/app.ts
```diff
@@ -26,12 +26,14 @@ import { load } from './server-settings';
 
   const port = process.env.PORT ? parseInt(process.env.PORT, 10) :  8531;
 
+  if (!process.env.PUBLIC_URI) {
+    process.env.PUBLIC_URI = 'http://localhost:' + port + '/';
+    console.log('PUBLIC_URI environment variable was not set, defaulting to *localhost*')
```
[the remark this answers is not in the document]
juhangsin: I'll fix this for the next PR!

PR #90, opened 2019-11-08
Reset password view

juhangsin:
This will let user to reset password if they lost their old password


src/reset-password/controller/redirect.ts
```diff
@@ -1,8 +1,38 @@
...
+import log from '../../log/service';
+import { EventType } from '../../log/types';
+import * as UserService from '../../user/service';
+import { User } from '../../user/types';
+import { resetPasswordForm } from '../formats/redirect';
 
-class ChangePasswordController extends Controller {
+class ResetPasswordController extends Controller {
 
+  async get(ctx: Context) {
```
evert: Both get and post should check if there is something inside ctx.state.session.resetPasswordUser and throw an error if there isnt

src/reset-password/controller/redirect.ts
```diff
@@ -1,8 +1,38 @@
...
+    const resetNewPassword = ctx.request.body.newPassword;
+    const confirmNewPassword = ctx.request.body.confirmNewPassword;
+
+    if (resetNewPassword !== confirmNewPassword) {
+      ctx.status = 303;
+      ctx.response.headers.set('Location', '/reset-password/change-password?msg=New+password+mismatch.+Please+try+again');
+      return;
+    }
+
+    await UserService.updatePassword(user, resetNewPassword);
```
evert: After updating the password, clear ctx.state.session.lostPasswordUser

src/reset-password/controller/token.ts
```diff
@@ -7,9 +7,8 @@ class ResetPasswordTokenController extends Controller {
   async get(ctx: Context) {
     const urlToken = ctx.state.params.token;
     const user = await validateToken(urlToken);
-    ctx.state.session.resetPassword = user.id;
-
-    ctx.status = 303;
+    ctx.state.session.resetPasswordUser = { user };
```
evert: Maybe the user doesn't need to be wrapped in another object

src/reset-password/controller/redirect.ts
juhangsin: Would you use .clear() function here or .remove() function?

evert: usually you can do `delete ctx.state.session.lostPasswordUser`

src/reset-password/controller/redirect.ts
```diff
@@ -1,8 +1,47 @@
...
+import * as UserService from '../../user/service';
+import { User } from '../../user/types';
+import { resetPasswordForm } from '../formats/redirect';
 
-class ChangePasswordController extends Controller {
+class ResetPasswordController extends Controller {
 
+  async get(ctx: Context) {
+    if (!ctx.state.session.resetPasswordUser) {
+      throw new Error('The "resetPasswordUser" property must be provided');
```
evert: Lets make this a `Forbidden` error. The message should be something that's a bit more user-focused.. maybe something like "You can only use this endpoint after you went to the 'forgot password' flow"

src/reset-password/controller/redirect.ts
```diff
@@ -1,8 +1,47 @@
...
+
+    if (resetNewPassword !== confirmNewPassword) {
+      ctx.status = 303;
+      ctx.response.headers.set('Location', '/reset-password/change-password?msg=New+password+mismatch.+Please+try+again');
+      return;
+    }
+
+    await UserService.updatePassword(user, resetNewPassword);
+    ctx.state.session.resetPasswordUser.clear()
+    ctx.state.session = {
```
evert: If you do this, you automatically log a user in. We don't really want that.

src/reset-password/controller/redirect.ts
```diff
@@ -1,8 +1,47 @@
...
+    ctx.response.body = resetPasswordForm(ctx.query.msg);
 
+  }
+
+  async post(ctx: Context) {
+    if (!ctx.state.session.resetPasswordUser) {
+      throw new Error('The "resetPasswordUser" property must be provided');
+    }
+
+    const user: User = ctx.state.session.resetPasswordUser.user;
```
evert: This line is no longer correct. It shouldn't be `.user` anymore because you removed the extra object.

src/reset-password/controller/redirect.ts
```diff
@@ -1,8 +1,45 @@
...
+    console.log('~~~~get~~~~', ctx.state.session.resetPasswordUser)
+    ctx.response.type = 'text/html';
+    ctx.response.body = resetPasswordForm(ctx.query.msg);
 
+  }
+
+  async post(ctx: Context) {
+
+    if (!ctx.state.session.resetPasswordUser) {
+      throw new Forbidden('You can only use this endpoint after you went to the \'forgot password\' flow');
```
evert:
```suggestion
      throw new Forbidden('You can only use this endpoint after you went through the \'forgot password\' flow');
```

src/reset-password/controller/redirect.ts
```diff
@@ -1,8 +1,45 @@
 import Controller from '@curveball/controller';
```
evert: Lets change the name of `redirect` to something else

juhangsin: Yeah, I'd like to sit down with you and talk about naming convention for few of them. because some of them doesn't represent what the method actually does

PR #91, opened 2019-11-09
Feature/reset password

juhangsin:


src/reset-password/controller/redirect.ts
```diff
@@ -0,0 +1,48 @@
+import Controller from '@curveball/controller';
```
evert: FIlenames should be corrected.. 'redirect' is not really what this controller does...

Issue #39, opened 2019-04-03
Add User-Agent to audit log


evert:
The audit log has an ip, but having a User-Agent would be really useful too.

PR #92, opened 2019-11-09
User agent

juhangsin:
#39


mysql-schema/027-userlog3.sql
```diff
@@ -0,0 +1,9 @@
+SET NAMES utf8mb4;
+START TRANSACTION;
+
+INSERT INTO changelog VALUES (27, UNIX_TIMESTAMP());
+
+ALTER TABLE user_log ADD
+  user_agent TEXT NOT NULL;
```
evert: This field should be nullable so previous entries already in the db don't fail

```suggestion
  user_agent TEXT NULL;
```

PR #95, opened 2019-11-11
Reset password redirect workaround

juhangsin:


src/reset-password/controller/redirect.ts
```diff
@@ -9,6 +9,12 @@ import { resetPasswordForm } from '../formats/redirect';
 
 class ResetPasswordController extends Controller {
 
+  /**
+   * Currently, this GET request isn't being used since there is a browser issue where
+   * it gets redirected from the email link, the browser clears the cookie
+   *
```
evert: We *do* actually use this endpoint, when a user submits something and the new passwords didn't match.

src/reset-password/service.ts
```diff
@@ -37,6 +47,9 @@ export async function sendResetPasswordEmail(user: User) {
     nodemailer.getTestMessageUrl(info);
 }
 
+/**
+ * Create unique token and stores in the reset_password_token table
```
evert: Grammar

src/reset-password/service.ts
```diff
@@ -49,6 +62,9 @@ export async function createToken(user: User): Promise<string> {
     return token;
 }
 
+/**
+ * Checks if the token provided is valid based on the time provided
```
evert: Not a very good description. There's no time provided to this function.

How about:

```
Checks if a 'password reset token' is valid, and returns the associated user.

This function only works once for every token. After calling this function, the token automatically gets deleted.
```

PR #107, opened 2019-11-15
User inactive fixes

### LINKED ISSUE #69: Throw a better error message when a user tries to log in but their account is not yet active.

juhangsin:
#69

### DIFF
--- a/src/user/service.ts
+++ b/src/user/service.ts
@@ -10,11 +10,12 @@ const fieldNames = [
   'nickname',
   'created',
   'type',
+  'active'
 ];
 
 export async function findAll(): Promise<User[]> {
 
-  const query = `SELECT ${fieldNames.join(', ')} FROM users WHERE active = 1`;
+  const query = `SELECT ${fieldNames.join(', ')} FROM users`;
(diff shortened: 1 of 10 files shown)


src/user/service.ts
```diff
@@ -39,7 +40,7 @@ export async function findById(id: number): Promise<User> {
 }
 export async function findByIdentity(identity: string): Promise<User> {
```
merrickread: With this function and `findAll()` above, it's generally best to pass the entire user object as a parameter since you can guarantee the typeset. I can show you examples in the media-api another day. 

Doesn't have to change for this PR, but something to keep in mind going forward.

src/user/types.ts
```diff
@@ -8,6 +8,7 @@ export type NewUser = {
   nickname: string,
   created: Date,
   type: UserType,
+  active: boolean
 };
```
merrickread: Also a nice way to declare the NewUser type is something like this:

`export type NewUser = Omit<User, 'id'>;`

It's the same type as the user, except you don't have an id yet. The id can be appended to the record once the record is inserted into the db.

Issue #106, opened 2019-11-14
Add support for group members.


evert:
This should be done after #105 

When we have groups in the system, these groups might be accessed via:

    /principal/123

If a user is a group, we should add the following API endpoint:

    /principal/123/member

This endpoint is a collection that is a list of users that are part of the group. The scope of this ticket is a read-only interface for this. We don't yet need to support adding members or removing them.

The implication of this ticket is also to add a new MySQL table such as 'group_members'.

evert:
cc @merrickread

PR #112, opened 2019-11-26
Support for group members

juhangsin:
#106
When the user's userType is a group (type = 3), a group will be able to see the group members when the endpoint is '/user/3/member'


mysql-schema/028-group-member.sql
```diff
@@ -0,0 +1,11 @@
+SET NAMES utf8mb4;
+START TRANSACTION;
+
+INSERT INTO changelog VALUES (28, UNIX_TIMESTAMP());
+
+CREATE TABLE group_members (
+  user_id INT UNSIGNED NOT NULL,
```
evert: Any MySQL table that you create should always have a primary key. In this case, a combination of user_id/group_id will only appear once so you should make a primary key with those 2 fields.

mysql-schema/028-group-member.sql
```diff
@@ -0,0 +1,11 @@
...
+  group_id INT UNSIGNED NULL
```
evert: Should also be `NOT NULL`

src/group/controller/collection.ts
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Cheks if user type is a group
+     */
+    const groupUser = await groupService.isGroup(user);
+
+    if (!groupUser) {
+      throw new BadRequest('User must be a group to gain access');
+    }
+
+    const groupMemebers = await groupService.findAllGroupMemebers(user);
```
evert: 2 typos

src/group/service.ts
```diff
@@ -0,0 +1,70 @@
...
+  'active',
+  'type'
+];
+
+/**
+ *  Checks if the user is a type group and returns true or false
+ */
+export async function isGroup(user: User): Promise<boolean> {
+
+  const query = 'SELECT type FROM users WHERE id = ? AND type = 3';
```
evert: You don't need to ask the database for the type, because the `user` parameter already has a `type` property.

src/group/types.ts
```diff
@@ -0,0 +1,3 @@
+import { User } from '../user/types';
+
+export type GroupMember = User;
```
evert: I don't think it's needed to make a new type for this. I actually think it's kinda helpful to know that the `groups` service deals with users, not a special `GroupMember`.

src/group/formats/hal.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function collection(user: User, group: GroupMember[]) {
+
+  const hal: any = {
+    _links: {
+      self: {
+        href: '/user/' + user.id + '/member',
+        title: user.nickname + ' members'
+      },
+      members: [],
```
evert: When a HAL resource is some kind of collection, the relationship type should always be `item`

src/group/service.ts
juhangsin: embarrassing :(

src/group/controller/collection.ts
juhangsin: embarrassing :(

src/group/types.ts
juhangsin: So would I just import the User from user/type.ts file and use the user instead?

juhangsin:
> Missing:
> 
> The response to GET /user/xyz should have a link somewhere to the `members` endpoint if it's a group

Would I be adding the `member` endpoint inside of user HAL or group HAL?

juhangsin:
> Missing:
> 
> The response to GET /user/xyz should have a link somewhere to the `members` endpoint if it's a group

So this is working now, I put the link inside of the

src/group/service.ts
```diff
@@ -0,0 +1,69 @@
+import database from '../database';
+import { User } from '../user/types';
+
+type GroupRecord = {
```
evert: I'm thinking there's still a lot of duplication here compared to the Users service. Maybe it's easier to just move the service functions in there?

src/group/service.ts
```diff
@@ -0,0 +1,69 @@
...
+  'nickname',
+  'created',
+  'active',
+  'type'
+];
+
+/**
+ *  Checks if the user is a type group and returns true or false
+ */
+export async function isGroup(user: User): Promise<boolean> {
```
evert: This function still has the same issue. Why does it need to do a MySQL query to get it's information

src/group/controller/collection.ts
```diff
@@ -0,0 +1,28 @@
...
+import * as hal from '../formats/hal';
+import * as groupService from '../service';
+
+class GroupMemberCollectionControlller extends Controller {
+  async get(ctx: Context) {
+
+    const user = await userService.findById(parseInt(ctx.state.params.id, 10));
+
+    /**
+     * Cheks if user type is a group
```
evert: Typo.

src/group/controller/collection.ts
```diff
@@ -0,0 +1,28 @@
...
+     */
+    const groupUser = await groupService.isGroup(user);
```
evert: The variable name suggests that you're getting something back that is a 'user object'

src/user/formats/hal.ts
```diff
@@ -40,6 +39,13 @@ export function item(user: User, privileges: PrivilegeMap) {
     type: TypeMap.get(user.type),
     privileges
   };
+
+  if (user.type === 3) {
+    hal._links.members = {
+      href: '/user/' + user.id + '/member',
+      title: 'Group Memebers'
```
evert: Typo

src/user/formats/hal.ts
```diff
@@ -40,6 +39,13 @@ export function item(user: User, privileges: PrivilegeMap) {
     type: TypeMap.get(user.type),
     privileges
   };
+
+  if (user.type === 3) {
```
evert: Is user.type ever tbe literal 3? I think it needs to be compared with ‘UserType.group‘

src/user/formats/hal.ts
```diff
@@ -40,6 +39,13 @@ export function item(user: User, privileges: PrivilegeMap) {
...
+    hal._links.members = {
```
evert: Usually when we link to a single resource that is a collection, the style we use would be something like 'member-collection' instead of 'members'.

I'm not sure how consistent we've been with this, but it makes spelling certain links less awkward, especially when the singular word already ends with and s.

src/group/service.ts
```diff
@@ -1,46 +1,27 @@
...
-  'active',
-  'type'
-];
+import * as UserService from '../user/service';
+import { User, UserType } from '../user/types';
 
 /**
  *  Checks if the user is a type group and returns true or false
  */
 export async function isGroup(user: User): Promise<boolean> {
```
evert: No longer needs to be async

src/group/service.ts
```diff
@@ -1,46 +1,27 @@
...
 
-  const query = `SELECT ${fieldNames.join(', ')} FROM users WHERE id = ? AND type = 3`;
-  const result = await database.query(query, [user.id]);
-
-  if (!result[0].length) {
+  if (user.type !== UserType.group) {
     return false;
   }
 
   return true;
```
evert: This should be a 1 liner

src/user/types.ts
```diff
@@ -20,3 +12,5 @@ export type User = {
   type: UserType,
   active: boolean
 };
+
+export type NewUser = Omit<User, 'id'>;
```
evert: :+1:

src/group/controller/collection.ts
```diff
@@ -11,11 +11,11 @@ class GroupMemberCollectionControlller extends Controller {
     const user = await userService.findById(parseInt(ctx.state.params.id, 10));
 
     /**
-     * Cheks if user type is a group
+     * Checks if user type is a group
      */
-    const groupUser = await groupService.isGroup(user);
+    const group = await groupService.isGroup(user);
```
evert: `group` still kinda suggests that the variable is a group itself, not a boolean. I would go for `isGroup`, or better yet... put this function call in the if statement itself. It also won't need `await` anymore

PR #114, opened 2019-12-03
Create user API 

juhangsin:


evert:
this new feature should return a `Conflict` error when a user already exists, and do validation

src/user/formats/hal.ts
```diff
@@ -1,12 +1,18 @@
...
-import { User, UserType } from '../types';
+import { NewUser, User, UserType } from '../types';
 
 const TypeMap = new Map<UserType, string>([
   [UserType.user, 'user'],
   [UserType.app, 'app'],
   [UserType.group, 'group'],
 ]);
 
+export const TypeMapInt = new Map<string, UserType>([
```
evert: This is a bit of an obscure name, and not super obvious what it does. That's kind of ok if it's not exported, but since this is now used by the controller, it might be better to change this up.

Maybe you can add a new function called `export function userTypeStringToUserType`. (still a bit of a weird name but i can't of anything better).

This function can be responsible for converting a string like `user` into the UserType enum

Issue #113, opened 2019-11-29
Support for invitations


juhangsin:
Currently we don't have invitation functionality in the system.

It needs to have 2 things:
1. When the non-existing user wants to have an account within the system, group owner/user will be able to send an invitation for the user to sign up
2. Group to be able to inviting an existing user who's not part of any group to be part of a specific group,

juhangsin:
Questions for @evert 

When the admin sends an invite email to a none existing user, would the URL contain a token or is it just a generic invite email with the `/register` link? Since what we have is anyone can register, I just want to make sure, but I understand that in the future or for a specific client, they would want this auth system to work as an invite-only.

evert:
I think starting off with a `/register` link is a pretty good start, but you are right.. this 100% needs to be tokenized for closed registration systems. I feel getting the initial flow right will help you make good steps in that direction though

juhangsin:
Current user being able to create a new non-user is completed.
Next step is a current user being able to send an email with the URI link for /register

PR #116, opened 2019-12-09
Invite user: basic HTML frontend

juhangsin:
#113


src/create-user/controller.ts
```diff
@@ -0,0 +1,43 @@
...
+      }
+    }
+
+    const newUser = await userService.save({
+      identity: ctx.request.body.identity,
+      nickname: ctx.request.body.nickname,
+      created: new Date(),
+      type: ctx.request.body.type,
+      active: 'active' in ctx.request.body
+    });
```
evert: validation

* identity must exist and must be a url
* nickname must have at least 1 character
* type must be 1 of the 3 allowed values

Issue #117, opened 2019-12-10
It should be possible for 'register' to be disabled and 'lost password' enabled


juhangsin:
project-pegasi/auth-api#9

juhangsin:
#118

PR #118, opened 2019-12-10
when register is disabled, it hides 'register' btn

juhangsin:
#117

### DIFF
--- a/src/login/controller.ts
+++ b/src/login/controller.ts
@@ -17,7 +17,7 @@ class LoginController extends Controller {
     ctx.response.body = loginForm(
       ctx.query.msg,
       ctx.query.error,
-      await getSetting('registration.enabled'),
+      (await getSetting('registration.enabled') === 'true'),
     );
 
   }
--- a/src/login/controller.ts
+++ b/src/login/controller.ts
@@ -12,12 +12,11 @@ import { loginForm } from './formats/html';
 class LoginController extends Controller {
 
   async get(ctx: Context) {
-
     ctx.response.type = 'text/html';
     ctx.response.body = loginForm(
       ctx.query.msg,
       ctx.query.error,
-      (await getSetting('registration.enabled') === 'true'),
+      await getSetting('registration.enabled'),
     );
 
   }
--- a/src/server-settings.ts
+++ b/src/server-settings.ts
@@ -8,14 +8,14 @@ export function getSetting(setting: string, deflt?: any): any {
     throw new Error('Settings have not been loaded. Call load() first');
   }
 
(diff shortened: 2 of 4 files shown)


src/login/controller.ts
```diff
@@ -17,7 +17,7 @@ class LoginController extends Controller {
     ctx.response.body = loginForm(
       ctx.query.msg,
       ctx.query.error,
-      await getSetting('registration.enabled'),
+      (await getSetting('registration.enabled') === 'true'),
```
evert: hmm why is this compared to a string?

juhangsin: when i changed the `registration.enabled` to false in `server-setting` table, it actually didn't recognize as a boolean, instead the type was a `string`, so i had to convert it

evert: I think the intent of the setting system is to treat the data as JSON, so it might mean that we're missing a `JSON.parse()` in the system that deals with getting settings from the database

juhangsin: Ah okay, so since the `getSetting` function takes the string and returns a string, i think that's why `true` is returning as a string. do you think it's bad idea to wrap that `await getSetting()` function in `JSON.parse()`, since there are other places that actually uses `getSetting()` function with params as a string?

evert: No, I think JSON.parse should be done for all settings when they are being loaded in.

juhangsin: so from `function loginForm(msg: string, error: string, registrationEnabled: boolean)`
the `registrationEnabled`'s value is a string `'true'` when i console log it. it's strange since we're explicitly setting the registrationEnabled's type as a boolean.

evert: getSetting() returns `any` and that basically causes typescript to completely stop type checking.

juhangsin: From the `getSetting()` func, would this be an 'okay' way to do?
```
if (value == 'true' || 'false') {
  return JSON.parse(value);
}
return value;
```

evert: No, JSON.parse should be called for every value

juhangsin: gotcha

src/server-settings.ts
```diff
@@ -8,7 +8,7 @@ export function getSetting(setting: string, deflt?: any): any {
     throw new Error('Settings have not been loaded. Call load() first');
   }
 
-  const value = settingsCache.get(setting);
+  const value = JSON.parse(settingsCache.get(setting));
```
evert: JSON.parse should be called when loading the settings from the database. Doing it there means JSON.parse only happens once, not every time getSetting is called.

PR #121, opened 2019-12-11
Lost password in on /authorize

juhangsin:
badgateway/pegasi-auth-api#8

### DIFF
--- a/src/oauth2/formats/html.ts
+++ b/src/oauth2/formats/html.ts
@@ -1,7 +1,7 @@
 import { render } from '../../templates';
 type KeyValue = { [key: string]: string };
 
-export function loginForm(msg: string, error: string, hiddenFields: KeyValue) {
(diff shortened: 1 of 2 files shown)


src/oauth2/formats/html.ts
```diff
@@ -1,13 +1,14 @@
...
 
-export function loginForm(msg: string, error: string, hiddenFields: KeyValue) {
+export function loginForm(msg: string, error: string, hiddenFields: KeyValue, registrationEnabled: boolean) {
 
   return render('login', {
     title: 'Login',
     msg: msg,
     error: error,
     hiddenFields: hiddenFields,
+    registrationEnabled,
```
nchangfong: nitpick: why is this argument the only one not named?

evert: the others can be cleaned up as well

juhangsin: I believe this is a new thing in ES6 where if key and value are the same, you can just directly call it without naming them?

PR #123, opened 2019-12-12
requried, optional, disable addition to the '2FA' field

juhangsin:
badgateway/pegasi-auth-api#10
HTML side done.
In the `server_settings`, depending on what the value is for `totp` setting, it will render different text

### DIFF
--- a/src/login/controller.ts
+++ b/src/login/controller.ts
@@ -53,8 +53,8 @@ class LoginController extends Controller {
       }
     } else if (await userService.hasTotp(user)) {
       return this.redirectToLogin(ctx, '', 'TOTP token required');
-    } else if (!ctx.request.body.totp && await getSetting('totp') === 'required') {
-      return this.redirectToLogin(ctx, '', 'TOTP code is required');
+    } else if (await getSetting('totp') === 'required') {
+      return this.redirectToLogin(ctx, '', 'The system administrator has made TOTP tokens mandatory, but this user did not have a TOTP configured. Login is disabled');
     }
 
     ctx.state.session = {
--- a/src/login/controller.ts
+++ b/src/login/controller.ts
@@ -17,6 +17,7 @@ class LoginController extends Controller {
       ctx.query.msg,
       ctx.query.error,
       await getSetting('registration.enabled'),
+      await getSetting('totp')
     );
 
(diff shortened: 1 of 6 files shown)


src/login/controller.ts
```diff
@@ -46,12 +47,14 @@ class LoginController extends Controller {
...
-        }
+      if (!await userService.validateTotp(user, ctx.request.body.totp)) {
+        log(EventType.totpFailed, ctx.ip(), user.id);
+        return this.redirectToLogin(ctx, '', 'Incorrect TOTP code');
+      }
     } else if (await userService.hasTotp(user)) {
-        return this.redirectToLogin(ctx, '', 'TOTP token required');
+      return this.redirectToLogin(ctx, '', 'TOTP token required');
+    } else if (!ctx.request.body.totp && await getSetting('totp') === 'required') {
+      return this.redirectToLogin(ctx, '', 'TOTP code is required');
```
evert: Hmm, i don't think this is correct yet. The error message also badly describes what the problem is.

I think a better error message might be:

   The system administrator has made TOTP tokens mandatory, but this user did not have a TOTP configured. Login is disabled.

I'm not sure if the new if statement captures this ...

templates/login.hbs
```diff
@@ -16,10 +16,18 @@
...
     </label>
 
+    {{#if totpEnabled}}
     <label for="totp">
+      {{#if totpRequired}}
+      <span>TOTP code  <em>(required)</em></span>
+      {{else}}
       <span>TOTP code  <em>(if set)</em></span>
+      {{/if}}
       <input type="text" name="totp" class="totp" id="totp" placeholder="Enter TOTP code" autocomplete="one-time-code" />
```
evert: Input field should get a `required="required"` attribute

src/login/controller.ts
juhangsin: Can you be a bit more specific why this is wrong? I have tested with the TOTP value to be "required", and when there isn't any value inside of the TOTP input, it will redirect to login with the error provided

evert: I *think* you can just drop the first half of the if statement.

It just looked a bit suspicious to me because it has this extra condition...

Issue #43, opened 2019-04-03
Add support for /.well-known/change-password


evert:
Some detail:

https://news.ycombinator.com/item?id=18618193

This only really makes sense after we have a "reset password" feature: #34

PR #130, opened 2019-12-20
add support for /.well-known/change-password

nchangfong:
closes: #43 

adds support for `/.well-known/change-password` and consolidates existing well-known routes in single folder.


evert:
Awesome, thank you @nchangfong

PR #134, opened 2019-12-30
Privilege controllers

evert:


mysql-schema/032-privilege-settings.sql
```diff
@@ -0,0 +1,13 @@
+SET NAMES utf8mb4;
+START TRANSACTION;
+
+INSERT INTO changelog VALUES (28, UNIX_TIMESTAMP());
+
+CREATE TABLE privileges (
```
evert: Every table _MUST_ have a primary key. In this case it should be the `privilege` field

PR #139, opened 2020-01-02
typo

juhangsin:


src/login/controller.ts
```diff
@@ -43,7 +43,7 @@ class LoginController extends Controller {
 
     if (!user.active) {
       log(EventType.loginFailedInactive, ctx.ip(), user.id, ctx.request.headers.get('User-Agent'));
-      return this.redirectToLogin(ctx, '', 'This account is inactiviated. Please contact Admin');
+      return this.redirectToLogin(ctx, '', 'This account is inactivated. Please contact Admin');
```
evert: I would say it's either deactivated or inactive, but not inactivated

Issue #135, opened 2019-12-30
Create a new 'privilege' collection


evert:
We need to have deeper support for privileges.

A good first step in that direction is to create a new 'privilege' collection.
This collection has a simple purpose. It: has a list of available privileges on the system.

This means we need a new `privilege` table. Each privilege should have the following fields:

1. privilege - A string. This can be a string such as 'admin' or 'read-users'.
2. description - Another string.

The following HTTP methods should be implemented:

    GET /privilege
    GET /privilege/admin

The purpose of this collection is mostly descriptional. It allows a user to read the list of privileges and see what kind of privileges are available on the system.

Tasks

* [x] Create new 'privileges` mysql table
* [x] Create the PrivilegeCollectionController class
* [x] Create the PrivilegeController class
* [x] Initialize the database by adding an `admin` privilege + description in the mysql schema, so we have something to look at.

The controllers should emit HAL.

Some work for this has been done in #134 

After this, in a separate PR it would be nice if the privilege collection also automatically lists any privileges that exist in `user_privileges`, but don't have a matching record in `privileges` yet.

We can chat about that once you're done this.

PR #140, opened 2020-01-02
including 'user_privilege' to privilege query

juhangsin:
related to #135

### DIFF
--- a/src/privilege/service.ts
+++ b/src/privilege/service.ts
@@ -1,5 +1,4 @@
 import { Context } from '@curveball/core';
-import { NotFound } from '@curveball/http-errors/dist';
 import db from '../database';
 import database from '../database';
 import { User } from '../user/types';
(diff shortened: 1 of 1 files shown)


src/privilege/service.ts
```diff
@@ -54,7 +53,7 @@ export async function hasPrivilege(who: User | Context, privilege: string, resou
 
 export async function findPrivileges(): Promise<Privilege[]> {
 
-  const query = 'SELECT privilege, description FROM privileges;';
+  const query = 'SELECT privilege, description FROM privileges UNION ALL SELECT DISTINCT scope, null FROM user_privileges LEFT JOIN privileges ON scope = privilege WHERE privilege IS NULL';
```
evert: Can you split this query over multiple lines?

juhangsin:
Question @evert.
So I made a mistake inside of sql file where the name of the file is 032-blah-blah, but in the change log number, since I copied and paste that one line, i forgot to change it
`INSERT INTO changelog VALUES (28, UNIX_TIMESTAMP());`
that 28 should've been 32, does it matter?

evert:
It _should_ be ok. Because you used a lower number, that patch would not have worked for anyone that tried to apply it.

Issue #40, opened 2019-04-03
Show 'country' in Audit Log API


evert:
We have an ip for every entry in the audit log, but it would be nice if we can automatically map this to a country.

This is only really a good idea if we can get a good free ip-to-country database we can integrate.

evert:
#209 replaces this PR

PR #151, opened 2020-03-02
map ip to country

claudiunicolaa:


claudiunicolaa:
resolves https://github.com/curveball/a12n-server/issues/40

mysql-schema/034-userlog4.sql
```diff
@@ -0,0 +1,9 @@
+SET NAMES utf8mb4;
+START TRANSACTION;
+
+INSERT INTO changelog VALUES (34, UNIX_TIMESTAMP());
+
+ALTER TABLE user_log ADD
+  country VARCHAR(50) NULL;
```
evert: Could we store the country as a 2-byte character string instead?

src/log/service.ts
```diff
@@ -67,3 +71,11 @@ function isContext(ctx: any): ctx is Context {
   return ((<Context> ctx).ip !== undefined);
 
 }
+
+function getCountryByIp(ip: string)
```
evert: I had a look in this library, and I think it's worth making a small adjustment.

The library will return a country, _or_ the string `?`. I don't think it's good to store `?` anywhere, so instead I think this function should have a condition for this and return `null` if a country could not be determined.

evert:
Also, thank you for the contribution. It's definitely appreciated!

evert:
This _might_ be a better alternative, as it packs in a database and is using a license that _is_ compatible with ours.

src/log/service.ts
```diff
@@ -2,6 +2,8 @@ import { Context } from '@curveball/core';
 import db from '../database';
 import { User } from '../user/types';
 import { EventType, LogEntry } from './types';
+// @ts-ignore: No types for this yet
```
evert: It looks like there is a `@types/geoip-lite` package, so this should not be needed when this is installed!

evert:
#209 replaces this PR

PR #160, opened 2020-03-03
Target es2019

juhangsin:
Changing target to es2019

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,7 +5,8 @@ Changelog
 0.12.5 (?)
 ----------
 
-* Changing target to es2019 to support optional chaining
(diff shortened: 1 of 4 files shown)


changelog.md
```diff
@@ -2,6 +2,11 @@ Changelog
 =========
 
 
+0.12.5 (?)
+----------
+
+* Changing target to es2019 to support optional chaining
```
evert:
```suggestion
* Typescript target is now `es2019` instead of `esnext` to ensure that older
  Node.js versions are supported.
```

evert:
Not needed

PR #164, opened 2020-03-04
Fix PUBLIC_URI bugs

### LINKED ISSUE #163: Process exits if PUBLIC_URI environment variable is not available
If the `PUBLIC_URI` is not available in the process environment the server should default to localhost + port. Currently the process exits if it is not set

evert:
* Better grammar
* Now correctly uses the default
* Added a ⚾

Fixes #163

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,10 +7,17 @@ import accessLog from '@curveball/accesslog';
 
 import { load } from './server-settings';
 
+const pkgInfo = require('../package.json');
+console.log('⚾ Curveball %s %s', pkgInfo.name, pkgInfo.version);
+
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -7,10 +7,17 @@ import accessLog from '@curveball/accesslog';
 
 import { load } from './server-settings';
 
+const pkgInfo = require('../package.json');
+console.log('⚾ Curveball %s %s', pkgInfo.name, pkgInfo.version);
```
merrickread: very hip

merrickread: ready for the big leagues

evert: <3

Issue #168, opened 2020-05-08
Implement PKCE


evert:
OAuth 2.1 will _require_ PKCE for the `authorization_code` grant_type going forward. We need to support this.

PKCE for authorization_code was originally defined here:

https://tools.ietf.org/html/rfc7636

But the OAuth2.1 specifcation will integrate that spec. Draft is here:

https://tools.ietf.org/html/draft-parecki-oauth-v2-1-01

I would say that the OAuth 2.1 draft is probably a better target than the RFC7636. The prose might be better although I think there's no differences.

PR #172, opened 2020-05-12
PKCE Support

mhum:
This is just a draft PR, but I think it's feature complete. I just want to add some more tests for the other parts of the code I touched.

The RFC gives the option of storing the code within the server or as part of the `code` that is returned from the authorization request. That seems weird to me, tbh. Since the authorization codes are short-lived in stored in our DB, I figured it was easiest just to store the code challenge, too.

### DIFF
--- a/src/oauth2/controller/authorize.ts
+++ b/src/oauth2/controller/authorize.ts
@@ -43,7 +43,7 @@ class AuthorizeController extends Controller {
     const responseType = ctx.query.response_type;
     const redirectUri = ctx.query.redirect_uri;
     const codeChallenge = ctx.query.code_challenge;
-    const codeChallengeMethod = ctx.query.code_challenge_method ? ctx.query.code_challenge_method : 'plain';
+    const codeChallengeMethod = ctx.query.code_challenge_method ? (<'S256'>ctx.query.code_challenge_method) : 'plain';
     const grantType = responseType === 'code' ? 'authorization_code' : 'implicit';
 
     try {
@@ -119,7 +119,7 @@ class AuthorizeController extends Controller {
     const redirectUri = ctx.request.body.redirect_uri;
     const responseType = ctx.request.body.response_type;
     const codeChallenge = ctx.request.body.code_challenge;
-    const codeChallengeMethod = ctx.request.body.code_challenge_method;
+    const codeChallengeMethod = ctx.request.body.code_challenge_method ? (<'S256'>ctx.request.body.code_challenge_method) : 'plain';
     const grantType = responseType === 'code' ? 'authorization_code' : 'implicit';
 
     try {
@@ -218,7 +218,7 @@ class AuthorizeController extends Controller {
     redirectUri: string,
     state: string|undefined,
     codeChallenge: string|undefined,
-    codeChallengeMethod: string|undefined,
+    codeChallengeMethod: 'S256' | 'plain' | undefined
   ) {
 
     const code = await oauth2Service.generateCodeForUser(
--- a/src/oauth2/controller/authorize.ts
+++ b/src/oauth2/controller/authorize.ts
@@ -29,13 +29,21 @@ class AuthorizeController extends Controller {
     if (!ctx.query.redirect_uri) {
       throw new InvalidRequest('The "redirect_uri" parameter must be provided');
     }
+    if (ctx.query.response_type === 'code') {
+      if (ctx.query.code_challenge_method && !['plain', 'S256'].includes(ctx.query.code_challenge_method)) {
+        throw new InvalidRequest('The "code_challenge_method" must be "plain" or "S256"');
+      }
+      if (!ctx.query.code_challenge && ctx.query.code_challenge_method) {
+        throw new InvalidRequest('The "code_challenge" must be provided');
+      }
+    }
     const clientId = ctx.query.client_id;
     const state = ctx.query.state;
     // const scope = ctx.query.scope;
     const responseType = ctx.query.response_type;
     const redirectUri = ctx.query.redirect_uri;
     const codeChallenge = ctx.query.code_challenge;
-    const codeChallengeMethod = ctx.query.code_challenge_method;
+    const codeChallengeMethod = ctx.query.code_challenge_method ? ctx.query.code_challenge_method : 'plain';
     const grantType = responseType === 'code' ? 'authorization_code' : 'implicit';
 
     try {
@@ -235,7 +243,7 @@ class AuthorizeController extends Controller {
   }
 
   /**
-   * We're overriding the default dipatcher to catch OAuth2 errors.
+   * We're overriding the default dispatcher to catch OAuth2 errors.
    */
   async dispatch(ctx: Context): Promise<void> {
 
--- a/src/oauth2/controller/authorize.ts
+++ b/src/oauth2/controller/authorize.ts
@@ -34,6 +34,8 @@ class AuthorizeController extends Controller {
     // const scope = ctx.query.scope;
     const responseType = ctx.query.response_type;
     const redirectUri = ctx.query.redirect_uri;
+    const codeChallenge = ctx.query.code_challenge;
+    const codeChallengeMethod = ctx.query.code_challenge_method;
     const grantType = responseType === 'code' ? 'authorization_code' : 'implicit';
 
     try {
@@ -61,7 +63,7 @@ class AuthorizeController extends Controller {
       if (responseType === 'token') {
         return this.tokenRedirect(ctx, oauth2Client, redirectUri, state);
       } else {
-        return this.codeRedirect(ctx, oauth2Client, redirectUri, state);
+        return this.codeRedirect(ctx, oauth2Client, redirectUri, state, codeChallenge, codeChallengeMethod);
       }
 
     } else {
@@ -73,6 +75,8 @@ class AuthorizeController extends Controller {
           state: state,
           redirect_uri: redirectUri,
           response_type: responseType,
+          code_challenge: codeChallenge,
+          code_challenge_method: codeChallengeMethod,
         },
         await getSetting('registration.enabled'),
         await getSetting('totp')
@@ -98,6 +102,8 @@ class AuthorizeController extends Controller {
     const state = ctx.request.body.state;
     const redirectUri = ctx.request.body.redirect_uri;
     const responseType = ctx.request.body.response_type;
+    const codeChallenge = ctx.request.body.code_challenge;
+    const codeChallengeMethod = ctx.request.body.code_challenge_method;
     const grantType = responseType === 'code' ? 'authorization_code' : 'implicit';
 
     try {
@@ -163,7 +169,7 @@ class AuthorizeController extends Controller {
     if (responseType === 'token') {
       return this.tokenRedirect(ctx, oauth2Client, params.redirect_uri, params.state);
     } else {
-      return this.codeRedirect(ctx, oauth2Client, params.redirect_uri, params.state);
+      return this.codeRedirect(ctx, oauth2Client, params.redirect_uri, params.state, codeChallenge, codeChallengeMethod);
     }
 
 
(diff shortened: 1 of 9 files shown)


src/oauth2/service.ts
```diff
@@ -142,7 +142,7 @@ export async function generateTokenForClient(client: OAuth2Client): Promise<OAut
  *
  * The resource owner then exchanges that code for an access and refresh token.
  */
-export async function generateTokenFromCode(client: OAuth2Client, code: string): Promise<OAuth2Token> {
+export async function generateTokenFromCode(client: OAuth2Client, code: string, codeVerifier: string|undefined): Promise<OAuth2Token> {
```
evert, line 145: I think it might be nice to make `codeVerifier` get a more specific type like `'S256' | 'plain'`, and make sure that this is validated as well. (earlier, in the controller)

src/oauth2/service.ts
```diff
@@ -169,6 +171,33 @@ export async function generateTokenFromCode(client: OAuth2Client, code: string):
...
+    return;
+  }
+
+  if (codeChallenge && !codeVerifier) {
+    // The authorization request started with PKCE, but the token request did not follow through
+    throw new InvalidRequest('The code verifier was not supplied');
+  }
+
+  // For the plain method, the derived code and the code verifier are the same
+  let derivedCodeChallenge = codeVerifier;
```
evert, line 186: Is `plain` secure? I don't really know the PKCE spec too well!

src/oauth2/service.ts
```diff
@@ -169,6 +171,33 @@ export async function generateTokenFromCode(client: OAuth2Client, code: string):
 
 }
 
+export function validatePKCE(codeVerifier: string|undefined, codeChallenge: string|undefined, codeChallengeMethod: string|undefined) {
```
evert, line 174: Could we simply *not* call this function when PKCE is not set up? I feel a bit better about that than to make every argument potentially undefined. This means that this check has to move to the call-site

src/oauth2/service.ts
mhum: Here, `codeVerifier` refers to the verification code the client sends in the body of the token request _after_ the authorization request happens (where the challenge method is provided). This is only ever a string (if provided) or undefined (if not provided).

But this does bring up the fact that I'm not validating the `codeChallengeMethod` in the authorization request. I did think of that later. I forgot to add _any_ validation to the authorization request (such as providing the `codeChallenge` but not the `codeChallengeMethod`). I'll add a check there to verify that the provided `codeChallengeMethod` (if included) is either `plain` or `S256`.

src/oauth2/service.ts
mhum: No, it's to accommodate clients that cannot do sha256 for technical reasons since PKCE is required in 2.1.

src/oauth2/service.ts
mhum: I guess I can. Feels a bit weird to me to have something like this:
```
if (codeVerifier !! codeChallenge !! codeChallengeMethod) {
  validatePKCE(codeVerifier, codeChallenge, codeChallengeMethod)
}

Since PKCE is required by 2.1 and supported in 2.0, every token request can potentially require PKCE validation. It makes more logical sense to me for that to be the first thing to do while doing PKCE validation (do we even have to do PKCE validation). I don't like separating the "do we need to do PKCE validation" and the "do PKCE validation" since it's potentially applicable to every request.

File: src/oauth2/service.ts (hunk 1 above)
Comment (evert) [reply src/oauth2/service.ts:145]: Cool!

File: src/oauth2/service.ts (hunk 2 above)
Comment (evert) [reply src/oauth2/service.ts:186]: I see. I remember one of the major issues with JWT was that many libraries by default also supported a 'none' encryption method. So I hope this is not a similar thing, but I would guess not.

File: src/oauth2/service.ts (hunk 3 above)
Comment (evert) [reply src/oauth2/service.ts:174]: I think the risk of also allowing undefined for any of the parameters is that it gets easier to slip in bugs, and call that function with a bad argument.

If the function requires a string, the caller is forced to ensure that it really is a string and if it isn't, get a hard error. I feel that this makes the source a bit more resilient against potential bugs.

You might also consider passing in something else, like an object with all the properties. An issue here too is that given each of the arguments may or may not be undefined, this creates 8 different branches, whereas a type like this:

```
type PkceChallenge = {
  codeVerifier: string,
  codeChallenge: string,
  codeChallengeMethod: 'plain' | 'S256'
} | undefined
```
Only creates 2 branches, and it's a bit more obvious what the behavior of the 2 branches will be.

File: src/oauth2/service.ts (hunk 2 above)
Comment (mhum) [reply src/oauth2/service.ts:186]: I would hope oauth libraries that support sha256 wouldn't provide the implementor the option of disabling it. The RFC does say the client MUST use sha256 if it can technically.
https://tools.ietf.org/html/draft-parecki-oauth-v2-1-01#section-4.1.1.2

File: src/oauth2/service.ts (hunk 3 above)
Comment (evert) [reply src/oauth2/service.ts:174]: Third alternative is to use overloads:

```typescript
 export function validatePKCE(codeVerifier: string, codeChallenge: string, codeChallengeMethod: string) 
 export function validatePKCE(codeVerifier: undefined, codeChallenge: undefined, codeChallengeMethod: undefined);
 export function validatePKCE(codeVerifier: string|undefined, codeChallenge: string|undefined, codeChallengeMethod: string|undefined) { 
```

This enforces that all arguments are either all undefined, or all strings.

mhum: If I don't make the parameters optional or provide an object where the fields are optional, I can't do validation if the user does something wrong. 

For example, the user provides a challenge code in the authorization request but doesn't provide the verifier code in the token request (in which case `codeVerifier` would be missing). Or if the user provides the verified code in the token request, but didn't provide the challenge code + method in the authorization request (so `codeChallenge` and `codeChallengeMethod` would be missing). 

I could push all of that parameter validation up into `generateTokenFromCode`, but that's a lot of logic specific to doing PKCE validation inside a function meant to generate tokens.

evert: Fair, that sounds good!

mhum:
This is ready for review. Changes since the initial draft:
- Param validation on authorization get/post requests
- Add a `start:watch` script for easy developing (by adding `tsc-watch`)
- Fixed a couple of `log` calls that were missing parameters
- Small spelling fix
- Added `chai-as-promised` for tests

src/oauth2/controller/authorize.ts
```diff
@@ -190,11 +212,20 @@ class AuthorizeController extends Controller {
...
   }
 
-  async codeRedirect(ctx: Context, oauth2Client: OAuth2Client, redirectUri: string, state: string|undefined) {
+  async codeRedirect(
+    ctx: Context,
+    oauth2Client: OAuth2Client,
+    redirectUri: string,
+    state: string|undefined,
+    codeChallenge: string|undefined,
+    codeChallengeMethod: string|undefined,
```
evert, line 221: Could this type be refined to `'S256' | 'plain' | undefined` vs just `string` ?

mhum: Is that valid syntax? I can't find anything in the TS docs about doing that. When I try it, I get an error on the calling side:
```
Argument of type 'string' is not assignable to parameter of type '"plain" | "S256"'.
```

mhum: Got it. Combo of string literals
https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types
and type assertions
https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions

evert: Basically, we want the strictest possible types for everything.

To fix the error, you also need to type the original definition the same way:

   let codeChallengeMethod: 'plain' | 'S256' | undefined;

It might also help to create a new reuable type for the enum. Ideally we use this everywhere this is applicable.

    type CodeChallengeMethod = 'plain' | 'S256';

The benefit if taking this approach and validating values *before* they enter the function, is that the function can only ever be called with a correct value, which means no validation in the function is needed.

I wrote a bit about this here:

https://evertpot.com/typescript-changes-code/

The official docs are here:

https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types

src/oauth2/controller/authorize.ts
```diff
@@ -43,7 +43,7 @@ class AuthorizeController extends Controller {
     const responseType = ctx.query.response_type;
     const redirectUri = ctx.query.redirect_uri;
     const codeChallenge = ctx.query.code_challenge;
-    const codeChallengeMethod = ctx.query.code_challenge_method ? ctx.query.code_challenge_method : 'plain';
+    const codeChallengeMethod = ctx.query.code_challenge_method ? (<'S256'>ctx.query.code_challenge_method) : 'plain';
```
evert, line 46: This is effectively a type cast, if the code_challenge_method is garbage, it will be typed as `S256`.

Ideal I think would be a strict check. Error when it's not one of the 2 acceptable values.

Bit more verbose, but:

```suggestion
let codeChallengeMethod: CodeChallengeMethod = undefined;
switch(ctx.query.code_challenge_method) {
  case: 'S256':
  case 'plain' :
    codeChallengeMethod = ctx.query.code_challenge_method;
    break;
  default:
    throw new Error('...');
}
```

src/oauth2/controller/authorize.ts
mhum: The validation still needs to happen _somewhere_ since these are user provided values. Even though TS will now prevent us from calling this function with a different value, validation now just needs to happen before calling the function.

src/oauth2/controller/authorize.ts
mhum: I'm effectively doing the same check on line 33:
https://github.com/curveball/a12n-server/blob/29e05d92d4594f352cff53cc23ffab4150bff79a/src/oauth2/controller/authorize.ts#L33

src/oauth2/service.ts
```diff
@@ -169,6 +171,33 @@ export async function generateTokenFromCode(client: OAuth2Client, code: string):
...
+
+  if (codeChallenge && !codeVerifier) {
+    // The authorization request started with PKCE, but the token request did not follow through
+    throw new InvalidRequest('The code verifier was not supplied');
+  }
+
+  // For the plain method, the derived code and the code verifier are the same
+  let derivedCodeChallenge = codeVerifier;
+
+  if (codeChallengeMethod === 'S256') {
```
evert, line 188: This is a suggestion, not a must... but if this were to be done with a switch (that also handles `plain`) you could use this approach:

https://stackoverflow.com/a/58009992/80911

to ensure that Typescript always complains if the enum ever gets expanded with a third "method".

The switch with `never` type is a bit ugly, so another option is to move this to a function that takes a `codeChallengeMethod`, a `codeVerifier` and returns the `derivedCodeChallenge`.

src/oauth2/controller/authorize.ts
evert: Maybe you can use an assertion function for easy re-use?

```
function assertIsChallengeMethod(input: string): asserts input is CodeChallengeMethod {

  if (input !== 'plain' && input !=='S256') {
    throw new Error('...');
  }
}
```

Used as such:

```
let codeChallengeMethod: CodeChallengeMethod | undefined;
if (ctx.query.code_challenge_method) {
   assertsIsChallengeMethod(ctx.query.code_challenge_method);
   codeChallengeMethod = ctx.query.code_challenge_method;
}
```

evert: An assertion function is 'trusted' by typescript to do the correct thing. Typescript will assume that you _will_ throw an error if the input was not correct.

After an assertion function is called, typescript will internally have narrowed down the `string` to `CodeChallengeMethod`.

mhum: I created a type and went with the switch statement.

src/oauth2/service.ts
mhum: I feel like a switch statement is overkill for only two possibilities. If there was a third type of challenge method, I'd create a separate function. I guess I could have done this with a ternary, too.

evert: The suggestion has more to related to getting a static error check once a third option is added. Basically when `CodeChallengeMethod` is altered with a new option, in the ideal case typescript refuses to compile it until that new case is explicitly handled. Switch is one way to handle this.. but this is not a big deal.

BTW I know other code in this repo is _also_ not really guarded as well and written this way.. I changed how I write things quite a bit since I wrote this stuff =)

Anyway, this works!

mhum:
Resolves #168

PR #174, opened 2020-07-05
Two-Step MFA Login

### LINKED ISSUE #26: Re-do the login form to make the 2FA token input a 'second step'
Right now the login form has 3 fields:

1. Username
2. Password.
3. TOTP (google authenticator code)

In the future we want to support the following cases as well:

1. TOTP is disabled, so there is no TOTP setup
2. Alternative forms of 2FA
3. WebAuthN (yubikey and such).

To do this effectively, what should happen is that the login screen just has a username + password field. After these are filled in, the user gets redirected and prompted for their TOTP code.

mhum:
I've added a second step to the login flow if TOTP is required or if the user as it enabled. I broke out the TOTP part of the login flow into a separate route/template. What it does on the login page is:
- If TOTP is disabled, just log the user in
- If the user has TOTP enabled, set a `mfa_user` in the `session` object and redirect the user to the next page

On the new `/mfa` route, it:
- Checks to see if `mfa_user` is set in the session. If not, send the user back to `/login`. This should prevent people from going directly to `/mfa`
- Does the same thing it used to: verifies the submitted token is correct and adds the `user` into the session

The above has implications for the `/authorize` route which is what leads to the largest change. Instead of duplicating yet more logic into `/authorize` and creating a new route like `/autorize_mfa`, I've updated the existing `/login` and the new `/mfa` to handle a `continue` query param that will redirect the user after logging in. If TOTP is disabled or the user doesn't have it set up, the user will be redirected to wherever `continue` points them. If they need to do the TOTP, they'll be first sent to `/mfa` and then redirected back to `/authorize` after supplying a valid token. 

For now, the `continue` param is verified to begin with `/authorize` as there is no reason to redirect the user anywhere else. In the future, I see this feature being useful when an un-authenticated user tries to hit a random page.We can bring them directly back after logging in. If we ever want to do that, the new `isValidateRedirect` function can be updated to use a set of whitelisted URLs the user can be redirected to.

I've tested `/login` and `/mfa` manually through the browser and `/authorize` with and without TOTP enabled through Postman.

Resolve #26


src/login/utilities.ts
```diff
@@ -0,0 +1,4 @@
+export function isValidateRedirect(url: string): boolean {
```
evert, line 1: I think this just needs to be called `isValidRedirect` but good otherwise!

mhum: Oops. I was going to call this `validateRedirect` at first and have it raise an Exception at first but decided to return a boolean instead.

evert:
Ty! This is really awesome.

PR #175, opened 2020-07-11
Yubikey Support

mhum:
Most of the logic for this change is completed. The new things are:
- Added an MFA page to the registration flow. In this PR, only adding a Yubikey is supported
- The MFA registration page adds a Yubikey device to the user as part of the registration flow
- When logging in (as long as MFA is enabled for the server), a user with MFA is presented with either the TOPT or the WebAuthn page

The changes to existing things are:
- I changed the existing MFA route from `/mfa` -> `/login/mfa` to better align the the new routes I added

What still needs to be done:
- DB migrations
- Clean up the front end JS to make the error handling more DRY and robust
- Generate random challenge when doing the attestation/assertion
- Figure out how to handle the [Relying Party Identifier](https://www.w3.org/TR/webauthn/#relying-party-identifier). The application needs to know what [expected origin](https://www.w3.org/TR/webauthn/#relying-party-identifier) is. This will either be an environmental variable or a server setting. I'm leaning towards server setting right now.

Nice followups to this PR:
- Supporting adding a TOTP device during registration
- Choose/fall-back to a different MFA device when logging in (for the case when WebAuthn doesn't work/isn't supported)

**Registration**
Registration Screen when MFA Registration setting is enabled
Screenshot from 2020-07-11 15-56-30

New MFA Registration screen
Screenshot from 2020-07-11 15-57-23

Begin Registration
Screenshot from 2020-07-11 15-57-54

Registration Completed
Screenshot from 2020-07-11 15-58-44

**Login**
Logging in with a Yubikey device
Screenshot from 2020-07-11 15-59-37

After Clicking Device Verification
Screenshot from 2020-07-11 15-59-57

**Incompatibilities**
This all works on my computer. However; I tested on my phone and with Postman and this **does not** work.

The request for the USB device isn't triggered. There must be something about using the embedded browser in Postman. I can't find anything when I Google.
Screenshot from 2020-07-11 16-03-47

On my phone, I just get an error.
Screenshot_20200711-160424


mhum:
If it would make reviewing easier, I can breaking this into two PRs: one for registering a new device and another for using it when logging in.

mhum:
I've decided to break this into two PRs. #177 is the first part.

PR #177, opened 2020-07-14
Register yubikey

mhum:
Adds an optional page (currently disabled) to the registration flow for registering a YubiKey. This PR will be followed up by another PR that verifies the device during log-in.

**Registration**
Registration Screen when MFA Registration setting is enabled
Screenshot from 2020-07-11 15-56-30

New MFA Registration screen
Screenshot from 2020-07-11 15-57-23

Begin Registration
Screenshot from 2020-07-11 15-57-54

Registration Completed
Screenshot from 2020-07-11 15-58-44


mysql-schema/035-webauthn-registration.sql
```diff
@@ -0,0 +1,21 @@
...
+  user_id INT UNSIGNED NOT NULL,
+  credential_id VARCHAR(2000) NOT NULL,
+  public_key VARCHAR(2000) NOT NULL,
+  counter INT UNSIGNED NOT NULL,
+  created INT UNSIGNED NOT NULL
+);
+
+INSERT INTO server_settings (setting, value) VALUES
+('registration.mfa.enabled', 'false'),
+('webauthn.relyingPartyId', '"localhost"'),
```
evert, line 17: Would it be possible to infer this? It will be a bit of a pain to have people update it.

mysql-schema/035-webauthn-registration.sql
```diff
@@ -0,0 +1,21 @@
...
+('webauthn.expectedOrigin', '"http://localhost:8531"'),
```
evert, line 18: Same here.

src/blob/controller.ts
```diff
@@ -13,6 +13,10 @@ class BlobController extends Controller {
         ctx.response.body = fs.readFileSync(__dirname + '/../../assets/' + ctx.state.params.filename);
         ctx.response.type = 'text/css';
         break;
+      case 'webauthn.js' :
```
evert, line 16: Really need to finish the `@curveball/static` package so we don't need this.

src/mfa/webauthn/types.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from '../../user/types';
+
+export type WebAuthnDevice = {
+    id: number,
+    user: User,
+    credentialID: string,
```
evert, line 6: Could this be `credentialId`?

src/register/controller/mfa.ts
```diff
@@ -0,0 +1,26 @@
...
+
+
+class MFAController extends Controller {
+
+  async get(ctx: Context) {
+    const user: User = ctx.state.session.registerUser;
+
+    if (!user) {
+      ctx.response.status = 303;
+      ctx.response.headers.set('Location', '/login');
```
evert, line 16:
```suggestion
      ctx.redirect(303, '/login');
```

evert:
Looks great. Clean too!

mysql-schema/035-webauthn-registration.sql
mhum: https://www.w3.org/TR/webauthn/#relying-party-identifier
> Note: An RP ID is based on a host's domain name. It does not itself include a scheme or port, as an origin does. The RP ID of a public key credential determines its scope. I.e., it determines the set of origins on which the public key credential may be exercised, as follows:
> - The RP ID must be equal to the origin's effective domain, or a registrable domain suffix of the origin's effective domain.
> - The origin's scheme must be https.
> - The origin's port is unrestricted.
> 
> For example, given a Relying Party whose origin is https://login.example.com:1337, then the following RP IDs are valid: login.example.com (default) and example.com, but not m.login.example.com and not com.
> 
> This is done in order to match the behavior of pervasively deployed ambient credentials (e.g., cookies, [RFC6265]). Please note that this is a greater relaxation of "same-origin" restrictions than what document.domain's setter provides.

No, this unfortunately cannot be inferred since the server won't know what domain it belongs to. Maybe this would work better with a different default value? 

Using the examples in this quote, `login.example.com` would be the Relying Party ID and `https://login.example.com:1337` would be the expected origin.

The `How does user registration work ?` section on https://www.okta.com/blog/2019/05/the-ultimate-guide-to-webauthn-registration-and-auth-flows/ explains this more simply.

mysql-schema/035-webauthn-registration.sql
mhum: See the explanation about the relying part id.

src/blob/controller.ts
mhum: ++

src/mfa/webauthn/types.ts
mhum: I did this so the type format would match the library. Specifically, its when doing this:
https://github.com/curveball/a12n-server/pull/178/files#diff-7acaf47777dd5c150438c29e0cd66842R50

`verifyAssertionResponse` 's single param is an object 
https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/assertion/verifyAssertionResponse.ts#L35
and the `authenticator` field is this type:
https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/assertion/verifyAssertionResponse.ts#L16
`AuthenticatorDevice` has a `credentialID` field
https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/typescript-types/src/index.ts#L105

I can definitely change it. `credentialID` is *not* consistent with the rest of the application. I was just being lazy.

mysql-schema/035-webauthn-registration.sql
evert: It's definitely going to be a bit of a pain to default to localhost, as this will not work out of the box for most people. I think in that case, it might be better to default to *no value* and report an error in the log that MFA will only be enabled once this value is set.

evert:
I think my only remaining concern at this point is the 'broken out of the box' issue. I wonder if we could get some good console.log descriptions that explains the issue. In the future being able to modify these settings in app would also be sweet :P but that's for later

evert:
Also great job on all the UI, looks really sweet

mysql-schema/035-webauthn-registration.sql
mhum: Yeah.... Definitely is a shame. It's why I defaulted `registration.mfa.enabled` to `false`. I'll set it to no value and log some errors as you suggested.

evert: Yea similar to the SMTP setting, I think we also default to *not* having a database value, and then echo something on the console along the lines off: "We're not sending email until you put something in the settings table for this"

PR #178, opened 2020-07-15
Verify fido device

mhum:
Verifies a YubiKey when logging in to the app or via the OAuth flow. When logging in, it will only show the MFA screen if the user has a TOTP or a WebAuthn device registered. I tested this in Chrome and Firefox as well as manually going through the OAuth flow outside of Postman (since Postman doesn't work with WebAuthn).

This is also the other half of #175. 

I also included one change from the existing functionality:
- I changed the existing MFA route from `/mfa` -> `/login/mfa` to better align the the new routes I added

Nice followups to this PR:
- Supporting adding a TOTP device during registration
- Choose/fall-back to a different MFA device when logging in (for the case when WebAuthn doesn't work/isn't supported)

**Login**
Logging in with a Yubikey device
Screenshot from 2020-07-11 15-59-37

After Clicking Device Verification
Screenshot from 2020-07-11 15-59-57

**Incompatibilities**
This all works on my computer. However; I tested on my phone and with Postman and this **does not** work.

The request for the USB device isn't triggered. There must be something about using the embedded browser in Postman. I can't find anything when I Google.
Screenshot from 2020-07-11 16-03-47

On my phone, I just get an error.
Screenshot_20200711-160424


mhum:
@evert This is the next PR in my chain for dependent PRs.

assets/webauthn.js
```diff
@@ -13,17 +14,19 @@ document.addEventListener('DOMContentLoaded', function(){
 
     let attResp;
     try {
-      attResp = await startAttestation(await resp.json());
+      const jsonResponse = await resp.json();
+      if (jsonResponse.status === 500) {
```
evert, line 18: Apologies if I don't fully understand this. Does `.status` exist on jsonResponse? Or should it be on `resp`?

It might also be better to capture a wider array of status codes. You could use `resp.ok`. I think we'd want to also handle other 4xx and 5xx codes?

assets/webauthn.js
```diff
@@ -39,11 +42,59 @@ document.addEventListener('DOMContentLoaded', function(){
...
+    elemError.innerHTML = '';
+    elemError.classList.add('hidden');
+    elemBeginLogin.disabled = true;
+
+    const resp = await fetch('/login/mfa/webauthn');
+
+    let asseResp;
+    try {
+      const jsonResponse = await resp.json();
+      if (jsonResponse.status === 500) {
```
evert, line 59: Potentially a similar problem

assets/webauthn.js
```diff
@@ -13,17 +14,19 @@ document.addEventListener('DOMContentLoaded', function(){
...
     try {
-      attResp = await startAttestation(await resp.json());
+      const jsonResponse = await resp.json();
+      if (jsonResponse.status === 500) {
+        handleError(jsonResponse.title, elemBeginRegister);
+      }
+      attResp = await startAttestation(jsonResponse);
     } catch (error) {
+      let errorText;
       if (error.name === 'InvalidStateError') {
```
evert, line 24: What emits this name? Should it be class based (instanceof) ?

mysql-schema/036-webauthn-setting.sql
```diff
@@ -0,0 +1,8 @@
+SET NAMES utf8mb4;
+START TRANSACTION;
+
+INSERT INTO changelog VALUES (36, UNIX_TIMESTAMP());
+
+INSERT INTO server_settings (setting, value) VALUES ('webauthn', '"disabled"');
```
evert, line 6: Alternatively, maybe use `false` instead of `"disabled"`. Or: `wabauthn.enabled` as a key?

assets/webauthn.js
mhum: `status` exists on the JSON response. It's coming from curveball/problem:
https://github.com/curveball/a12n-server/blob/master/src/main-mw.ts#L19
https://github.com/curveball/problem/blob/master/src/index.ts#L69

I'll make the error handling more robust.

assets/webauthn.js
mhum: ++

assets/webauthn.js
mhum: It's from the spec that browser implement: https://www.w3.org/TR/webauthn/#createCredential
I'm assuming the browser throws this error when the browser's API is called here:
https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/browser/src/methods/startAttestation.ts#L37

I took it from the package's example: https://github.com/MasterKale/SimpleWebAuthn/blob/34935e25bb3b2c25d005c653670b9c517ff8b050/example/public/register/index.html#L46

mysql-schema/036-webauthn-setting.sql
mhum: I was following the existing pattern used with totp:
https://github.com/curveball/a12n-server/blob/master/mysql-schema/030-totp-setting.sql#L6

Valid options are `disabled`, `optional`, and `required`.

I wouldn't be opposed to adding two fields: `enabled` and `required` where both are booleans.

assets/webauthn.js
mhum: No longer using the `status` field and just doing `resp.ok` as suggested since it handles all the 5XX and 4XX codes. I also added some error handling if the fetch request itself fails.

mysql-schema/036-webauthn-setting.sql
evert: fair, lets keep it!

PR #180, opened 2020-07-18
Add TOTP Device During Registration

mhum:
Add support for registering a TOTP device when creating a user account. There are two new UI pages:
- After registering, choose what kind of MFA device to add
- Adding the TOTP device

The new `totp.serviceName` setting is for setting what the service name is called when saved to a TOTP authenticator app.

I tested this with both Google Authenticator and Authy on Android with both the QR code and entering the secret manually.

Some refactors to existing code:
- Moved the WebAuthn AJAX endpoints from `/register/mfa/webauthn` -> `/register/mfa/webauthn/attestation` as `/register/mfa/webauthn` is required for the device setup page
- Moved and renamed the registration-related templates into a new folder called `register`

In a follow-up PR, I'll move the existing TOTP code out of `user/service` and into the new `mfa/totp/service` module I created as part of this PR.

MFA device selection screen
Screenshot from 2020-07-18 12-28-19

TOTP device registration screen
Screenshot from 2020-07-18 12-29-05


src/mfa/webauthn/controller/attestation.ts
```diff
@@ -0,0 +1,95 @@
+import Controller from '@curveball/controller';
```
mhum, line 1: This is just copying the content of `src/mfa/webauthn/controller/register.ts` into this new file.

src/mfa/webauthn/service.ts
```diff
@@ -6,25 +6,25 @@ import { User } from '../../user/types';
...
+  public_key: string;
+  counter: number;
+};
 
 export async function save(webAuthNDevice: NewWebAuthnDevice): Promise<WebAuthnDevice> {
   if (!isExistingDevice(webAuthNDevice)) {
     const query = 'INSERT INTO user_webauthn SET ?, created = UNIX_TIMESTAMP()';
 
-    const newUserRecord: Partial<UserWebAuthnRow> = {
+    const newWebAuthnRecord: Partial<UserWebAuthnRow> = {
```
mhum, line 20: Fixing a copypasta error from an earlier PR

mhum:
@evert This is the last review in my series

src/mfa/webauthn/controller/attestation.ts
```diff
@@ -0,0 +1,95 @@
...
+    const user: User = ctx.state.session.registerUser;
+    const body = ctx.request.body;
+
+    const expectedChallenge = ctx.state.session.webAuthnChallengeRegister;
+    ctx.state.session.webAuthnChallengeRegister = null;
+
+    if (!getSetting('webauthn.expectedOrigin')) {
+      console.error(`Error: The webauthn.expectedOrigin server-setting is not set. This should match the origin the browser is making the request from (ex: https://login.example.com)
+See the Relying Party Identifier section of the WebAuthn W3C Recommendation here:
+https://www.w3.org/TR/webauthn/#rp-id`);
```
evert, line 54: What's expected to happen if this setting is not set? Can we still continue?

mhum: If this is not set or the setting is incorrect, the server will return an error to the client. It still "works", but clients will receive errors until the correct origin is set. The server is expected to validate that the origin of the client that is sending the request matches the origin that it expects. This is to prevent clients from registering devices from origins the client doesn't expect.

The client error looks like this when registering a device:
Screenshot from 2020-08-13 21-26-51
`http://bogus-localhost:8531` is the server setting.

This setting is also used for logging in (assertion). The client gets a similar error:
Screenshot from 2020-08-13 21-20-16

mhum: It's number 2 in Step 6 from `Registration` described here:
https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API

evert: Ah I guess I just got confused because there was no `throw` or `return`..

Awesome!

PR #176, opened 2020-07-14
Tslint to eslint

juhangsin:

PR #181, opened 2020-08-03
Add typescript type checking back to lint target

mhum:
When `tslint` was removed in #176, it also removed Typescript type checking during the testing step. This adds it make to the `lint` make target. 

I could also make a new `check-types` make target and have that be ran as part of the `test` npm script if you all think that would be more appropriate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ test:
 	nyc mocha
 
 lint:
+	npx tsc --noemit
 	npx eslint --quiet 'src/**/*.ts' 'test/**/*.ts'
 
 fix:


evert:
Actually this is a bit different from all our other repos. I think we just run `make lint test` in ci environments, and `make watch` during dev. Don't really have anything against this though...

juhangsin:
> Actually this is a bit different from all our other repos. I think we just run `make lint test` in ci environments, and `make watch` during dev. Don't really have anything against this though...

@evert So this doesn't impact other repos (ie, `new-package`)?

evert:
No, I don't really think we need a separate rule in our packages unless peeps think this is really useful

mhum:
The reason I added it back was so Travis would do the type checking in the PR. This protects against changes made outside of running the app in watch mode (merging branches, writing code against tests only, etc.).

evert:
Are we not running `make test` in travis? That's probably the more important one =)

mhum:
It looks like it's actually `npm test` that is ran, but close enough. The reason for this change is so Travis will do type checking again since eslint doesn't do that.

evert:
Sorry I wasn't super clear. I assumed that the tests would run `tsc` and therefore this would already work... but it looks like that might actually not be the case, so lets merge this!

Issue #20, opened 2019-03-14
OAuth2 revoke token support


evert:
https://tools.ietf.org/html/rfc7009

evert:
@juhangsin , take a look at the RFC and see if you think you can do it

PR #182, opened 2020-08-07
Revoke endpoint

mhum:
New endpoint for revoking access/refresh tokens. The current implementation always revokes the access/refresh token pair.

I'm a bit confused on how to interrupt the first part of the following from the end of section 2.1. For the second part, that will be true since refresh and access tokens are stored in the same row.
> If the particular
   token is a refresh token and the authorization server supports the
   revocation of access tokens, then the authorization server SHOULD
   also invalidate all access tokens based on the same authorization
   grant (see Implementation Note).  If the token passed to the request
   is an access token, the server MAY revoke the respective refresh
   token as well.

Let me use an example to see if my interruption is correct. I request 2 tokens using the authorization code grant type and then 1 with the client credentials. If I try and revoke 1 of my refresh tokens I got from the authorization code grant, I should also revoke the _other_ access/refresh token pair from the authorization grant? And the access/refresh token pair from the client credentials grant remains?

This paragraph is the rationale for revoking the access/refresh pair together regardless of which is passed. I'm just not sure if I should be revoking _all_ of the access tokens when a refresh token is passed.

Resovle #20

### DIFF
--- a/src/log/types.ts
+++ b/src/log/types.ts
@@ -7,6 +7,7 @@ export enum EventType {
   resetPasswordSuccess,
   loginFailedInactive,
   webAuthnFailed,
+  tokenRevoked,
   oauth2BadRedirect = 11,
 }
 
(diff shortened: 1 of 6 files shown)


mhum:
@evert Do you have any thoughts on the interpretation of that paragraph from the RFC before I merge this?

evert:
I missed the description of this PR, but my interpretation is that, since our access/refresh tokens are always in a pair, one invalidates the other. I think in our case we just always want to destroy both the access and refresh token when either is specified.

Maybe we'll want something more specific down the road, but this will satisfy the main requirement well enough, which is a definite log out.

I don't think we want to log users out of every session though, so only the directly associated access/refresh token should be destroyed.

PR #183, opened 2020-08-13
dotenv

juhangsin:


juhangsin:
is it okay to have password for mysql to be @ dotenv file?
```
MYSQL_PASSWORD=password
MYSQL_USER=a12nserver
MYSQL_DATABASE=a12nserver
```

mhum:
I think it's okay since these are the defaults and it is set up in such a way that a real deployment can provide their own credentials. The "better" way would be to continue using environmental variables (which this doesn't prevent people from doing).

It would probably be nice to have a big, fat, scary warning in the logs on startup if the default values are still being used.

juhangsin:
@mhum & @evert, is it okay to merge this in?

mhum:
shipit!

PR #190, opened 2020-09-30
Better redirect_uri messages

evert:


src/oauth2/service.ts
```diff
@@ -54,6 +54,26 @@ export async function validateRedirectUri(client: OAuth2Client, redirectUrl: str
...
 
+/**
+ * Checks if a redirect_uri is permitted for the client.
+ *
+ * If not, it will emit an InvalidGrant error 
+ */
+export async function requireRedirectUri(client: OAuth2Client, redirectUrl: string): Promise<void> {
+
+  const query = 'SELECT id, uri FROM oauth2_redirect_uris WHERE oauth2_client_id = ?';
+  const result = await db.query(query, [client.id, redirectUrl]);
```
juhangsin, line 65: query only requires for `oauth2_client_id`, and you're passing in `client.id` & `redirectUrl`, is there reason for this?

evert: Yea that's wrong. Good eye

PR #194, opened 2020-10-08
dotenv doc

juhangsin:
Wondering if these also needs to change
```
Docker:
```sh
export MYSQL_PASSWORD=....
export MYSQL_USER=username
export MYSQL_DATABASE=databasename
docker run -it --rm -p 127.0.0.1:8531:8531 --name a12n-server-01 a12n-server
```

Not docker:
```sh
export MYSQL_PASSWORD=....
export MYSQL_USER=username
export MYSQL_DATABASE=databasename
make start
```
```


evert:
dotenv is not used in docker, so those environment variables are relevant. I doubt now though that just running `export FOO=bar` actually will have an effect on docker.

evert (approved):

PR #198, opened 2020-10-25
Allow users to add OAuth2 clients via the Web 

evert:
Unfortunately has some unrelated changes from updated eslint rules. srry


src/oauth2-client/controller/collection.ts
```diff
@@ -17,11 +20,65 @@ class ClientCollectionController extends Controller {
...
     ctx.response.body = hal.collection(user, clients);
 
   }
 
+  async post(ctx: Context) {
+
+    const user = await userService.findById(ctx.state.params.id);
+    if (!await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('Only users with the "admin" privilege can inspect OAuth2 clients that are not your own');
+    }
```
mhum, line 33: I've seen this duplication of access control logic in a few of the controllers. It would be nice if there was a method per-controller that could be applied to _all_ methods in the controller. Then access control on that resource could be centralized and lower the possibility of forgetting about it.

This isn't at all blocking on this review; it's just an observation.

src/oauth2-client/controller/collection.ts
```diff
@@ -17,11 +20,65 @@ class ClientCollectionController extends Controller {
...
+    const allowedAllowedGrantTypes: GrantType[] = [
+      'password',
+      'client_credentials',
+      'refresh_token',
+      'implicit',
+      'authorization_code'
+    ];
+
+    for(const agt of allowedGrantTypesArr) {
+      if (!allowedAllowedGrantTypes.includes(agt)) {
```
mhum, line 60: nit: This seems like a great use for Array's `.every()`

src/oauth2-client/controller/collection.ts
```diff
@@ -17,11 +20,65 @@ class ClientCollectionController extends Controller {
...
+
+    const redirectUris = ctx.request.body.redirectUris.split(' ');
+
+    if (!clientId) {
+      clientId = randomId(10);
+    } else if (clientId.length < 6) {
+      throw new UnprocessableEntity('clientId must be at least 6 characters or left empty');
+    }
+
+    const clientSecret = randomId(20);
```
mhum, line 45: nit: Doesn't really matter, but it doesn't look like we need this variable until line 69.

src/oauth2-client/service.ts
```diff
@@ -100,9 +100,39 @@ export async function getOAuth2ClientFromBody(ctx: Context): Promise<OAuth2Clien
 
 }
 
+export async function create(client: Omit<OAuth2Client, 'id'>, clientSecret: string, redirectUris: string[]): Promise<OAuth2Client> {
+
+  const query = 'INSERT INTO oauth2_clients SET ?';
+  const params = {
+    client_id: client.clientId,
+    client_secret: await bcrypt.hash(clientSecret, 12),
```
mhum, line 108: Maybe I'm missing something, but does `clientSecret` need to be a param to this function? Should this value be available at `client.clientSecret`?

src/oauth2-client/controller/collection.ts
evert: Yeah totally agreed. Would be nice to have a controller-wide way to handle this. Was also thinking of doing something with decorators.

src/oauth2-client/service.ts
evert: This is definitely wrong

src/oauth2-client/controller/collection.ts
mhum: Decorators would probably work, too. I've used them with other languages/frameworks. I keep forgetting that decorators are in draft proposal for JS.

mhum (approved):
Awesome! This will be super handy.

PR #200, opened 2020-10-28
dotenv is not a dev dep

evert:


juhangsin (approved):
shoot, my bad. thanks for fixing this! :+1:

PR #205, opened 2020-11-14
Stronger typing for server settings.

evert:
I think this needs a larger overhaul and unify how environment
variables, server settings and defaults are handled.

But this is a small step there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: node_js
 node_js:
   - stable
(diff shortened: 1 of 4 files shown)


mhum (approved):
Nice! One of the tests is broken. That stub can be deleted as the `getSettings` call isn't used in that class anymore. I also saw one other places that has `await getSettings` that can can be cleaned up, too.

PR #214, opened 2020-12-14
fix for `continue` query params.

juhangsin:
`ctx.query.continue` doesn't contain anything, but `ctx.request.body.continue` still contains the redirect URL.


evert (approved):
Makes perfect sense!

PR #217, opened 2020-12-16
Redis Session

merrickread:
If `REDIS_SESSION` is true, use `redis-session` for session management, otherwise default to memory store.

`REDIS_SESSION` default is false

### DIFF
--- a/src/main-mw.ts
+++ b/src/main-mw.ts
@@ -19,7 +19,7 @@ export default function(): Middleware {
     }),
     problem(),
     session({
-      store: process.env.REDIS_SESSION ? new RedisStore({
+      store: process.env.REDIS_HOST ? new RedisStore({
         prefix: 'A12N-session',
         clientOptions: {
           port: process.env.REDIS_PORT ? parseInt(process.env.REDIS_PORT, 10) : 6379,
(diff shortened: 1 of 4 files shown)


src/main-mw.ts
```diff
@@ -18,7 +19,13 @@ export default function(): Middleware {
     }),
     problem(),
     session({
-      store: 'memory',
+      store: process.env.REDIS_SESSION ? new RedisStore({
+        prefix: 'A12N-session',
```
mihok, line 23: Might be worthwhile to have this as a env var as well

mihok (approved):
LGTM

src/main-mw.ts
```diff
@@ -18,7 +19,13 @@ export default function(): Middleware {
...
+        clientOptions: {
+          port: process.env.REDIS_PORT ? parseInt(process.env.REDIS_PORT, 10) : 6379,
+          host: process.env.REDIS_HOST ? process.env.REDIS_HOST : '127.0.0.1',
+        },
+      }) : 'memory',
```
evert, line 28: Can you remove the `REDIS_SESSION` environment variable, and just use the existence of `REDIS_HOST` as a way to determine we're going to use Redis?

evert (changes requested):
Looks good otherwise!

src/main-mw.ts
merrickread: yep!

evert (approved):

src/main-mw.ts
evert: agreed, could be helpful if multiple environments use the same redis instance. Maybe something like REDIS_PREFIX ?

merrickread:
evert for merging this PR is there anything I need to do regarding versioning?

PR #220, opened 2021-01-25
Managing a group's members list

juhangsin:


src/group/controller/collection.ts
```diff
@@ -23,4 +23,4 @@ class GroupMemberCollectionControlller extends Controller {
```
evert, line 23: The most important endpoint is probably the `put` on this resource, letting us update the list of members. This is more consistent with everything else. The form is also nice though, as it lets us easily make changes.

juhangsin:
As suggested, I'm gonna focus on the API call for now, and I'll add the proper form after.

src/group/service.ts
```diff
@@ -30,3 +32,22 @@ export async function findAllGroupMembers(user: User): Promise<User[]> {
   return models;
 
 }
+
+export async function findByUserId(userId: number): Promise<any> {
```
evert, line 36: This method duplicates `findAllGroupMembers` . Why make another?

src/group/service.ts
```diff
@@ -30,3 +32,22 @@ export async function findAllGroupMembers(user: User): Promise<User[]> {
...
+  const query = 'SELECT * FROM group_members WHERE user_id = ?';
+  const result = await database.query(query, [userId]);
+
+  if (result[0].length !== 1) {
+    throw new NotFound(`Can't find group member with User ID ${userId}`);
+  }
+
+}
+
+export async function save (userId: string, group: User): Promise<void> {
```
evert, line 47: I would expect a `save` function in a group save a group. This does something very different.

src/group/service.ts
```diff
@@ -30,3 +32,22 @@ export async function findAllGroupMembers(user: User): Promise<User[]> {
...
+
+  const query = `INSERT INTO group_members SET group_id = ${group.id}, user_id = ?`;
```
evert, line 49: SQL injection

src/group/controller/collection.ts
evert: This has not been addressed yet. We usually update resources with `PUT` and the request body should match what we got with `GET`

src/group/controller/collection.ts
```diff
@@ -14,13 +15,39 @@ class GroupMemberCollectionControlller extends Controller {
...
+      if (!(err instanceof NotFound)) {
+        throw err;
+      }
+    }
+
+    await groupService.save(userBody.userId, group);
+
+    ctx.response.status = 201;
+    ctx.response.headers.set('Location', '/user/' + group.id + '/member');
+  }
```
evert, line 48: This suggests that a new resource has been created with this location, and that's not really true.

src/group/type.ts
```diff
@@ -0,0 +1,4 @@
+export type GroupMember = {
```
juhangsin, line 1: i guess this type isn't being used at all, so we could remove it

evert: Removed!

juhangsin (commented):
:+1:

PR #224, opened 2021-01-26
Some updates to OAuth2 'new client' form.

evert:
Hopefully this will reduce friction a bit.

Screenshot from 2021-01-26 00-27-54


juhangsin (approved):
this is gonna be super convenient!

PR #226, opened 2021-01-27
Fix OAuth2 error handler.

evert:
OAuth2 has its own error format, so needs its own error handling layer.
This existed before, but broke at some point, causing `undefined` error messages related to oauth2

This fixes error handling once more, as a new middleware.

### DIFF
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -31,13 +31,14 @@ import oauth2Metadata from './well-known/controller/oauth2-metadata';
 import clients from './oauth2-client/controller/collection';
 import client from './oauth2-client/controller/item';
 import clientNew from './oauth2-client/controller/new';
+import oauth2ErrorHandler from './oauth2/oauth2-error-handler';
 
 const routes = [
   router('/', home),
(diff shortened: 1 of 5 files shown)


src/routes.ts
```diff
@@ -31,13 +31,14 @@ import oauth2Metadata from './well-known/controller/oauth2-metadata';
...
+import oauth2ErrorHandler from './oauth2/oauth2-error-handler';
 
 const routes = [
   router('/', home),
   router('/assets/:filename', blob),
 
-  router('/authorize', oauth2Authorize),
-  router('/token', oauth2Token),
+  router('/authorize', oauth2ErrorHandler, oauth2Authorize),
+  router('/token', oauth2ErrorHandler, oauth2Token),
```
evert, line 41: This is a new router feature. Specifying multiple middlewares on the router level.

juhangsin (approved):

src/routes.ts
mhum: That's pretty cool!

mhum (approved):

PR #231, opened 2021-02-05
accepts full url instead of only rel url.

juhangsin:
i feel like splitting against `PORT` might be a bad idea, but using port instead of `PUBLIC_URI` env varible because userHref requres this exact format
`/user/:userId`, and sometimes, for the `PUBLIC_URI` env varible, user might or might not add the ending `/` which case it would be inconstant.

Let me know if there's a better solution to this! :)


src/group/controller/collection.ts
```diff
@@ -68,19 +68,26 @@ class GroupMemberCollectionController extends Controller {
...
-    if (ctx.request.body.userHref === undefined) {
+    if (requestBody.userHref === undefined) {
       throw new UnprocessableEntity('The "userHref" property must be set, and must point to the member you are adding to the group.');
     }
 
-    const newMemberHref = ctx.request.body.userHref;
-    let newMember;
+    let newMemberHref;
+    if (ctx.request.body.userHref.includes(process.env.PUBLIC_URI)) {
+      newMemberHref = requestBody.userHref.split(process.env.PORT).pop();
```
evert, line 85: 100% this is bad. There is no port in the url in most production environments. Use `new URL()` to parse the url and extract the information you need.

I also think handling urls is actually done in the services, so it's weird that the logic for this piece of this in the controller. I think it just needs to be done in `userService` (getByHref or getByUri) so it works magically everywhere.

src/user/service.ts
```diff
@@ -274,3 +274,16 @@ function isExistingUser(user: User | NewUser): user is User {
...
+  const matches = href.match(/^\/user\/([0-9]+)$/);
+
+  if (!matches) {
+    const userRelUrl = new URL(href).pathname;
+    return userRelUrl;
+  } else {
+    return href;
+  }
+
+}
```
evert, line 289: This is a very strange function. It validates /user/123 for one path, doesn't do what the name says, is async but doesn't need to be

src/group/controller/collection.ts
```diff
@@ -78,11 +78,13 @@ class GroupMemberCollectionController extends Controller {
...
     }
 
+
     const newMemberHref = ctx.request.body.userHref;
     let newMember;
 
     try {
-      newMember = await userService.findByHref(newMemberHref);
+      newMember = await userService.getByHref(newMemberHref);
+      newMember = await userService.findByHref(newMember);
```
evert, line 87: The logic of getByHref should be part of findByHref. It's also not great to re-use variables for different purposes.

But the idea is if i call `findByHref`, I expect to be able to to pass a url (absolute or relative).

src/user/service.ts
```diff
@@ -274,3 +273,12 @@ function isExistingUser(user: User | NewUser): user is User {
...
 }
+
+function getByHref(href: string): string {
+
+  const fullUrl = new URL(href);
+  if (fullUrl) {
+    return fullUrl.pathname;
+  }
+  return href;
+}
```
evert, line 284: Looking better!

The name is still a little weird though... `getByHref` .... get 'what' by href?  Since it's a users service, the name suggest to me it's basically the same as findByHref... try to think about what this actually does.

Second, could you write a unit test for this? We don't have a lot of those in a12nserver, but it would be nice to see some proof that this function fully works.

juhangsin:
My first attempt for writing tests.. Let me know if you think there should be more test cases.

src/user/service.ts
```diff
@@ -274,11 +274,15 @@ function isExistingUser(user: User | NewUser): user is User {
 
 }
 
-function getByHref(href: string): string {
+export function getRelHref(href: string): string {
```
evert, line 277: `Rel` generally in our hypermedia apis stands for `Relation` not `Relative`. How about getPathName?

test/test.ts
```diff
@@ -1,5 +1,26 @@
...
+    it('should pass with relative url being passed in', () => {
+      const url = '/user/1';
+      const result = getRelHref(url);
+      expect(result).to.equal(expectedValue);
+    });
+
+    it('should fail if URL doesn`t contain the opening /', () => {
+      const url = 'user/1';
+      const result = getRelHref(url);
+      expect(result).to.not.equal(expectedValue);
```
evert: Usually if a test is described as something like 'should fail', I think my assumption is that an exception is thrown.

However, it here it seems that should fail' is interpreted as 'return the wrong result'. That doesn't sound right to me

test/test.ts
```diff
@@ -1,5 +1,26 @@
-describe('this package', () => {
-  it('should have tests', () => {
-    // Write your test
-  });
+import { getRelHref } from '../src/user/service';
```
evert, line 1: Move this file to `test/user/service.ts`

(note that there's already other tests in this pacakge)

test/test.ts
```diff
@@ -1,5 +1,26 @@
...
+import { expect } from 'chai';
+
+describe('URI Test', () => {
```
evert, line 4: The description should probably just be the name of the function you are testing.

Then when you run the test, the output will say "getPathName should pass with a full url passed in"

evert (approved):

PR #235, opened 2021-02-22
one time token for user invitation

juhangsin:
adding one-time-token for user invitation


src/one-time-token/controller.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as userService from '../user/service';
+import { User } from '../user/types';
+import { createToken } from '../reset-password/service';
+
+class OneTimeTokenController extends Controller {
+
+  async post(ctx: Context<any>) {
+
+    if (!await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('Only users with the "admin" privilege can create new users');
```
evert, line 14: wrong error message.

It might also be nice at this point to start making more specific privileges for specific tasks? We can do that in another ticket though.

src/one-time-token/controller.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    let user: User;
+    try {
+      user = await userService.findByIdentity(ctx.state.user.identity);
+    } catch (err) {
+      if (err instanceof NotFound) {
+        ctx.redirect(303, '/reset-password?error=Account+not+found.+Please+try+again');
```
evert, line 22: In the case of a NotFound error I think you really just want to return a 404 and not redirect to reset password for this endpoint.

src/one-time-token/controller.ts
```diff
@@ -0,0 +1,39 @@
...
+        return;
+      } else {
+        throw err;
+      }
+    }
+
+    const token = await createToken(user);
+    const url = process.env.PUBLIC_URI + 'reset-password/token/' + token;
```
evert, line 30: returning the URL with the token sounds like a good idea to me, but a good way to return a url is with a hal link.

src/reset-password/service.ts
```diff
@@ -48,10 +48,10 @@ export async function sendResetPasswordEmail(user: User) {
 }
 
 /**
- * This function will create a unique token then store it in the databse
+ * This function will create a unique token then store it in the database
  */
 export async function createToken(user: User): Promise<string> {
-  const token = crypto.randomBytes(32).toString('base64').replace(/+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
+  const token = crypto.randomBytes(32).toString('base64').replace(/'+'/g, '-').replace(/\//g, '_').replace(/=+$/, '');
```
evert, line 54: What's going on here?

evert (commented):
Missing: link from user resource.

src/reset-password/service.ts
juhangsin: This was throwing an error
`SyntaxError: Invalid regular expression: /+/: Nothing to repeat`

evert: Well you fixed the error but the regex is wrong

src/one-time-token/controller.ts
```diff
@@ -0,0 +1,39 @@
...
+      throw new Forbidden('Only users with the "admin" privilege can request for one time token');
+    }
+
+    let user: User;
+    try {
+      user = await userService.findByIdentity(ctx.state.user.identity);
+    } catch (err) {
+      if (err instanceof NotFound) {
+        ctx.status = 404;
+        return;
```
evert, line 24: We don't set the status for errors, we throw exceptions. In this case you don't need to because the `NotFound` exception was already thrown. So all you have to do is *not* catch it.

src/one-time-token/controller.ts
```diff
@@ -0,0 +1,39 @@
...
+      } else {
+        throw err;
+      }
+    }
+
+    const token = await createToken(user);
+    const url = new URL(process.env.PUBLIC_URI + 'reset-password/token/' + token);
```
evert, line 31: Using the URL object makes sense, specifically for resolving relative urls. if you want to do that, don't concatenate the path and domain but use the different arguments of the URL constructor.

src/one-time-token/formats/hal.ts
```diff
@@ -0,0 +1,9 @@
+export function oneTimeToken(url: URL) {
+  return {
+    _links: {
+      self: {
+        href: url
```
evert, line 5: self should refer to the URL of the resource, so whatever url was used to access the endpoint.

Make a new link for the lost-password link.

In a previous version you also had the token itself in the body, and i think that that might still be helpful!

src/reset-password/service.ts
```diff
@@ -48,10 +48,10 @@ export async function sendResetPasswordEmail(user: User) {
 }
 
 /**
- * This function will create a unique token then store it in the databse
+ * This function will create a unique token then store it in the database
  */
 export async function createToken(user: User): Promise<string> {
-  const token = crypto.randomBytes(32).toString('base64').replace(/+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
+  const token = crypto.randomBytes(32).toString('base64').replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
```
evert, line 54: nice!

evert (changes requested):

evert:
I'll make these changes because i wanna test the release process with github actions

PR #241, opened 2021-02-23
one-time token bug fix

juhangsin:
fixing entire one-time token object was being passed in for token uri. 
also includes `ctx.state.user.identity` is the frontend user info instead of the new user we're trying to create.


evert (approved):
ty. weird that typescript didn't warn !

juhangsin:
> ty. weird that typescript didn't warn !

I think this is worth to fix for future PR's, I'll create an issue or put it in my TODO list.

PR #242, opened 2021-02-23
Add 'login.defaultRedirect'

juhangsin:


src/login/controller/login.ts
```diff
@@ -73,8 +73,20 @@ class LoginController extends Controller {
...
+      ctx.response.headers.set('Location', '/');
+    }
+
+  }
 
+  isRedirect() {
+    if (getSetting('login.defaultRedirect') !== '/') {
+      return true;
+    }
+    return false;
```
evert, line 89:
```suggestion
    ctx.response.redirect(303, getSetting('login.defaultRedirect', '/'));
```

evert: I think this will work

src/server-settings.ts
```diff
@@ -10,6 +10,7 @@ type Settings = {
   'webauthn.expectedOrigin': string,
   'webauthn.relyingPartyId': string,
   'webauthn.serviceName': string,
+  'login.defaultRedirect': '/' | string,
```
evert, line 13: just make this `string`. Not a bad idea but `/` is still just a string

evert (changes requested):

src/login/controller/login.ts
juhangsin: ahhhh yeah, i guess function has default params :( my bad

evert (approved):

PR #270, opened 2021-03-12
Update webauthn dependency.

evert:


src/mfa/webauthn/controller/attestation.ts
```diff
@@ -13,13 +13,16 @@ class WebAuthnAttestationController extends Controller {
...
     const attestationOptions = generateAttestationOptions({
-      serviceName: getSetting('webauthn.serviceName'),
+      rpName: getSetting('webauthn.serviceName'),
       rpID: getSetting('webauthn.relyingPartyId', new URL(process.env.PUBLIC_URI!).host),
       userID: user.id.toString(),
       userName: user.nickname,
       timeout: 60000,
-      attestationType: 'direct',
-      excludedCredentialIDs: (await webAuthnService.findDevicesByUser(user)).map(device => device.credentialID),
+      attestationType: 'indirect',
```
mhum, line 21: I know this is recommended by the package's docs, but I'm not sure the full implications. The spec isn't super clear either:
https://www.w3.org/TR/webauthn-2/#dom-attestationconveyancepreference-indirect
https://www.w3.org/TR/webauthn-2/#anonymization-ca

I guess using indirect and anonymizing the data means we would know less about the user (like the device they're using?). I'm not sure what parts of the attestation is anonymize-able. 

Yubico's docs recommend direct:
https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/WebAuthn_Client_Registration.html

mhum (approved):

src/mfa/webauthn/controller/attestation.ts
evert: So, what's kind of interesting is if i use the yubikey with firefox, firefox will actually give me the option to anonymize the data. (But then it also warns that not everyone will accept it).

It sounds like the benefit of 'direct' is that we can potentially remove devices from the system in the future if they are found out to be insecure.

Given the amount of people using this (not that many), I'm ok with being on the side of letting more people use the system vs. having this ability.

PR #278, opened 2021-03-25
adding change pw without current pw

juhangsin:


src/user/controller/password.ts
```diff
@@ -0,0 +1,46 @@
...
+import { User } from '../types';
+import * as privilegeService from '../../privilege/service';
+
+class UserPasswordController extends Controller {
+
+  async put(ctx: Context) {
+
+    if (!await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('Only users with the "admin" privilege may create new users');
+    }
```
juhangsin, line 14: from our conversation
`need a new API that allows anyone with admin privileges to set someone's password without knowing the old one`

but I feel that our purpose (NMC) of this endpoint is when the invited user is going through the `onboarding` process, they'll type in their desired password, but the user actually won't have the `admin` access for this.

so maybe removing this privilege check is necessary?
thoughts?

evert: No, this check is perfect. We don't want users to change their own passwords.

So you need 'admin' access to do this right now. And one of the things will have admin access, is the nmc api backend. The backend authenticates as itself to do this

src/user/controller/password.ts
```diff
@@ -0,0 +1,46 @@
...
+    let user: User;
+    try {
+      user = await UserService.findByIdentity('mailto:' + userBody.identity) as User;
+    } catch (err) {
+      if (err instanceof NotFound) {
+        throw new NotFound('User email doesn\'t exist');
+      } else {
+        throw err;
+      }
+    }
```
evert, line 27: The route for this is `/user/id/password`. You should get the user from the path, not the body.

evert, line 27: What is the purpose of this catch block?

src/user/controller/password.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    const userNewPassword = userBody.newPassword;
+    const confirmNewPassword = userBody.confirmNewPassword;
```
evert, line 30: This is an API, so it doesn't really need to confirm the password with a second field. The API would just copy the field anyway.

src/user/controller/password.ts
```diff
@@ -0,0 +1,46 @@
...
+
+
+    if (userNewPassword !== confirmNewPassword) {
+      throw new Unauthorized('New password mismatch, please try again');
+    }
+
+    await UserService.updatePassword(user, userNewPassword);
+
+    ctx.response.status = 201;
+    ctx.response.headers.set('Location', `/user/${user.id}`);
```
evert, line 40: Can you explain the purpose of these 2 lines?

evert (changes requested):

src/user/controller/password.ts
juhangsin: so my thought was making sure the user will type the correct email address during onboarding survey process. and when the email address doesn't match the database, it throws an error, i'll change this.

src/user/controller/password.ts
juhangsin: i thought what this does is after `updatePassword` successfully passes, the status code will be 400 -> 201, then redirect you to `/user/:userId`?

src/user/controller/password.ts
evert: Yes but it already threw an error before too =)

src/user/controller/password.ts
evert: No, `201` is specifically for creating new resources, and when there is a Location header with a `201` it tells the client : We just created this resource at this uri.

So the way this is done right now, after changing a password the API says "I just created a user".

evert: So `Location` has 2 uses:

1. Redirects (all 3xx codes)
2. To tell a client where a new resource got created (only 201)

src/user/controller/password.ts
```diff
@@ -14,30 +14,12 @@ class UserPasswordController extends Controller {
...
-
-    if (userNewPassword !== confirmNewPassword) {
-      throw new Unauthorized('New password mismatch, please try again');
-    }
 
     await UserService.updatePassword(user, userNewPassword);
 
-    ctx.response.status = 201;
-    ctx.response.headers.set('Location', `/user/${user.id}`);
+    return ctx.redirect(303, `/user/${user.id}`);
```
evert, line 22: Because this is 'just an API' and not for browser consumption, we don't really need a redirect. Just a `204` is fine here.

src/user/controller/password.ts
```diff
@@ -0,0 +1,28 @@
...
+class UserPasswordController extends Controller {
+
+  async put(ctx: Context) {
+
+    if (!await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('Only users with the "admin" privilege may create new users');
+    }
+
+    const userBody: any = ctx.request.body;
+    const user = await UserService.findByHref(ctx.params.href) as User;
```
evert, line 17: I don't think you have a `href` here. It's also better to not cast, because in this case it might be wrong. What if you got an app or a group?

src/user/controller/password.ts
```diff
@@ -0,0 +1,28 @@
...
+    const userNewPassword = userBody.newPassword;
+
+    await UserService.updatePassword(user, userNewPassword);
```
evert, line 20: Could we just call it `password` ?

src/user/controller/password.ts
```diff
@@ -0,0 +1,28 @@
...
+class UserPasswordController extends Controller {
+
+  async put(ctx: Context) {
+
+    if (!await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('Only users with the "admin" privilege may create new users');
+    }
+
+    const userBody: any = ctx.request.body;
+    const user = await UserService.findById(parseInt(ctx.params.id, 10)) as User;
```
evert, line 17: Last thing I think you missed. `findById` can return type `User`, `App` or `Group`.

By casting to `User` you are basically saying: "I am going to pretend I got a User back", even if you didn't. Don't cast, and instead throw (for example) a `Forbidden` error, because you should not be able to set a password for Apps or Groups.

evert (approved):

src/user/controller/by-href.ts
```diff
@@ -13,6 +13,7 @@ class UserByHrefController extends Controller {
 
     let hasControl = false;
     let hasPassword = false;
+    let hasPrivilege = false;
```
evert, line 16: `hasPrivilege` is not a good description because there are many types of privileges. `isAdmin` might be more appropriate here.

src/user/controller/by-href.ts
```diff
@@ -21,12 +22,16 @@ class UserByHrefController extends Controller {
     if (hasControl && user.type === 'user') {
       hasPassword = await userService.hasPassword(user);
     }
+    if (privilegeService.hasPrivilege(user, 'admin', '*')) {
```
evert, line 25: Check out line 19 for a good way to call this. This returns a promise, and because there is a database call it would be better if we only run this function once per request.

evert (changes requested):

src/user/controller/by-href.ts
evert, line 16: Would suggest:

```typescript
const isAdmin = await privilegeService.hasPrivilege(ctx, 'admin');
```

Now you have a single variable that can be re-used later in this function too.

src/user/controller/by-href.ts
```diff
@@ -13,25 +13,24 @@ class UserByHrefController extends Controller {
 
     let hasControl = false;
     let hasPassword = false;
-    let hasPrivilege = false;
+    const isAdmin = await privilegeService.hasPrivilege(ctx, 'admin');
+
     if (ctx.state.user.id === user.id) {
       hasControl = true;
     } else if (await privilegeService.hasPrivilege(ctx, 'admin')) {
```
evert, line 20: This line should use the new flag instead of doing the same operation again, so we do fewer database queries.

src/user/controller/item.ts
```diff
@@ -13,7 +13,8 @@ class UserController extends Controller {
 
     let hasControl = false;
     let hasPassword = false;
-    let hasPrivilege = false;
+    const isAdmin = await privilegeService.hasPrivilege(ctx, 'admin');
+
     if (ctx.state.user.id === user.id) {
       hasControl = true;
     } else if (await privilegeService.hasPrivilege(ctx, 'admin')) {
```
evert, line 20: Same here

evert (changes requested):

evert (approved):

PR #284, opened 2021-03-31
Edit user through HAL forms

mihok:
### Changes

Ive added the ability for users (if admin) to edit users (and themselves.)

This is the first PR, next will be providing a JSON textarea to update user permissions

### DIFF
--- a/src/user/formats/hal.ts
+++ b/src/user/formats/hal.ts
@@ -140,18 +140,6 @@ export function edit(user: Principal): HalResource {
             },
             value: user.active ? 'true' : 'false',
           },
-          {
-            name: 'type',
-            prompt: 'Type',
-            options: {
-              inline: [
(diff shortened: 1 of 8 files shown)


src/user/formats/hal.ts
```diff
@@ -84,17 +84,80 @@ export function item(user: Principal, privileges: PrivilegeMap, hasControl: bool
...
+            name: 'type',
+            prompt: 'Type',
+            options: {
+              inline: [
+                { prompt: 'User', value: 'user' },
+                { prompt: 'App', value: 'app' },
+                { prompt: 'Group', value: 'group' },
+              ],
+            },
+            value: user.type,
```
evert, line 153: It shouldn't be possible to change the type, because this can have all kinds of weird effects.

evert (approved):

evert (approved):

PR #285, opened 2021-03-31
Privilege policy editor

mihok:
### Changes

This creates a privilege policy editor form that folks cen directly edit a privilege policy JSON document.

### DIFF
--- a/src/privilege/service.ts
+++ b/src/privilege/service.ts
@@ -89,6 +89,18 @@ export async function findPrivilege(privilege: string): Promise<Privilege> {
 
 }
 
+export async function replacePrivilegeForUser(user: Principal, privilegeMap: PrivilegeMap): Promise<void> {
+  const query = 'DELETE FROM user_privileges WHERE user_id = ?';
+  await database.query(query, [ user.id ]);
+
+  for (const [ resource, privileges ] of Object.entries(privilegeMap)) {
+        for (const privilege of privileges) {
+                await addPrivilegeForUser(user, privilege, resource);
+        }
+  }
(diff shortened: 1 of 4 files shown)


src/privilege/service.ts
```diff
@@ -89,6 +89,18 @@ export async function findPrivilege(privilege: string): Promise<Privilege> {
...
+  await database.query(query, [ user.id ]);
+
+  for (const [ resource, privileges ] of Object.entries(privilegeMap)) {
+    for (const privilege of privileges) {
+      await addPrivilegeForUser(user, privilege, resource);
+    }
+  }
+
+}
+
```
evert, line 103: Multiple queries should be wrapped in a transaction

src/user/controller/privileges.ts
```diff
@@ -0,0 +1,48 @@
...
+import * as hal from '../formats/hal';
+import * as userService from '../service';
+// import * as groupService from '../../group/service';
+
+class UserEditPrivilegesController extends Controller {
+
+  async get(ctx: Context) {
+
+    const user = await userService.findById(+ctx.params.id);
+    const privileges = await privilegeService.getPrivilegesForPrincipal(user);
```
evert, line 15: The issue with this is that it doesn't just pull in a users' immediate privileges, but also all the privileges from any groups the user is in.

So I think we need a new function in the privilege service (like `getDirectPrivileges` or `getImmediatePrivileges`)

evert (changes requested):

evert (approved):

PR #290, opened 2021-04-04
JWT support for OAuth2 access tokens

evert:
Defined here: https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt-12

If JWT_PRIVATE_KEY is set and contains a RS256 private key, access
tokens will be generated as JWTs.

More related features to this will follow, but this is the first step to
JWT support.


src/oauth2/controller/active-sessions.ts
```diff
@@ -38,4 +44,11 @@ class UserActiveSessions extends Controller {
 
 }
 
+function assertNotGroup(input: Principal): input is User | App {
```
juhangsin, line 47:
```suggestion
function assertNoneGroup(input: Principal): input is User | App {
```

wouldn't this make more sense?

src/oauth2/controller/active-sessions.ts
```diff
@@ -20,6 +22,10 @@ class UserActiveSessions extends Controller {
       throw new Forbidden('You can only use this API for yourself yourself, or if you have \'admin\' privileges');
     }
 
+    if (!assertNotGroup(user)) {
```
juhangsin, line 25:
```suggestion
    if (!assertNoneGroup(user)) {
```
Wouldn't this make more sense?

juhangsin (approved):
just a name change (i think), but otherwise, looks great!

src/oauth2/controller/active-sessions.ts
evert: It asserts that 'input' is _not_ a group. Input can be either a user, group or app but it's always just 1.

An alternative could have been `assertsAppOrUser` but it's a bit more clumsy to write. Or `assertsPrincipalHasOAuthSupport` but it's getting a bit wordy then :/

PR #292, opened 2021-04-05
Markdown home document

evert:
This PR adds an alternative format to the `/` API endpoint, rendered in markdown.

### DIFF
--- a/src/home/controller.ts
+++ b/src/home/controller.ts
@@ -1,16 +1,40 @@
-import Controller from '@curveball/controller';
+import Controller, { accept, method } from '@curveball/controller';
 import { Context } from '@curveball/core';
 import hal from './formats/hal';
+import markdown from './formats/markdown';
 import * as privilegeService from '../privilege/service';
+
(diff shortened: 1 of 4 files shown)


src/home/controller.ts
```diff
@@ -1,16 +1,40 @@
...
+  @method('GET')
+  @accept('application/hal+json')
+  async getJson(ctx: Context) {
 
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const version = require('../../package.json').version;
     const user = ctx.state.user;
     const isAdmin = await privilegeService.hasPrivilege(user, 'admin', '*');
+    ctx.response.headers.append('Link', [
+      '</>; rel="alternate"; type="text/markdown"',
```
mihok, line 36: What is this empty element thing here?

mihok (approved):
LGTM

src/home/controller.ts
evert: In a Link header the uri goes between `< `and `>`. So the path is just `/`

PR #300, opened 2021-04-15
Allowing users to be enabled via the one-time-token API

evert:


src/one-time-token/controller/exchange.ts
```diff
@@ -26,6 +33,16 @@ class OneTimeTokenExchangeController extends Controller {
     }
 
     const user = await tokenService.validateToken(ctx.request.body.token);
+
+    if (!user.active) {
+      if (ctx.request.body.activateUser) {
+        user.active = true;
+        await userService.save(user);
+      } else {
+        throw new Forbidden('The user associated with the one-time-token has been deactivated. Either activate the user first, or provide the "activate" property in the request if the intent is to activate the user with the one-time-token mechanism');
```
BeckyPollard, line 42: 🆒

BeckyPollard (approved):

PR #302, opened 2021-04-20
Create new password for users. 

juhangsin:
This PR is for invited user to be able to create a password.

### DIFF
--- a/src/user/controller/password.ts
+++ b/src/user/controller/password.ts
@@ -3,8 +3,34 @@ import { Context } from '@curveball/core';
 import * as UserService from '../service';
 import { Forbidden } from '@curveball/http-errors';
 import * as privilegeService from '../../privilege/service';
+import log from '../../log/service';
+import { EventType } from '../../log/types';
 
 class UserPasswordController extends Controller {
+
+  async post(ctx: Context<any>) {
+
+    if (!await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('Only users with the "admin" privilege may create new password');
+    }
+
+    const user = await UserService.findById(parseInt(ctx.params.id, 10));
+    const userNewPassword = ctx.request.body.newPassword;
+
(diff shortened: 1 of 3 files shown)


src/user/controller/password.ts
```diff
@@ -3,9 +3,35 @@ import { Context } from '@curveball/core';
...
+    if (user.type !== 'user') {
+      throw new Forbidden('You can only create passwords for users');
+    }
+
+    await UserService.createPassword(user, userNewPassword);
+
+    ctx.session = {
+      user: user,
+    };
+    log(EventType.changePasswordSuccess, ctx);
```
juhangsin, line 29: Should I create `EventType.createPasswordSuccess` for this?

evert: I don't really think this endpoint is needed. `PUT` can be used for creating or modifying, and we don't really want the client to have to decide whether it should be created or modified. The only thing really needed to fix this bug is to change the `UPDATE` query to `INSERT ... ON DUPLICATE KEY UPDATE`

evert:
I don't really think this endpoint is needed. PUT can be used for creating or modifying, and we don't really want the client to have to decide whether it should be created or modified. The only thing really needed to fix this bug is to change the UPDATE query to INSERT ... ON DUPLICATE KEY UPDATE.

`POST` is not strictly for creation. We actually prefer `PUT` for creating resources _if_ the client can know the URI of the newly created resource.

juhangsin:
> I don't really think this endpoint is needed. PUT can be used for creating or modifying, and we don't really want the client to have to decide whether it should be created or modified. The only thing really needed to fix this bug is to change the UPDATE query to INSERT ... ON DUPLICATE KEY UPDATE.
> 
> `POST` is not strictly for creation. We actually prefer `PUT` for creating resources _if_ the client can know the URI of the newly created resource.

BeckyPollard (approved):
This works great Ju! 🙂

PR #311, opened 2021-04-29
Stats for principals

juhangsin:


src/principal/service.ts
```diff
@@ -27,6 +27,22 @@ export async function findAll(): Promise<Principal[]> {
 
 }
 
+export async function getPrincipleStats(): Promise<PrincipalStats> {
+
+  const query = 'SELECT type, COUNT(*) as total FROM principals GROUP BY type';
+  const result = await database.query(query);
+
+  const principalStats: any = {};
+  // const principalStats: Record<PrincipalType, number> = {};
```
juhangsin, line 36: Is there way to type this object?

evert: Type it as `PrincipalStats` and instead of starting with an empty object, start with an object with a 3 properties and a value of `0` for each.

src/user/controller/collection.ts
```diff
@@ -16,7 +16,8 @@ class UserCollectionController extends Controller {
   async get(ctx: Context) {
 
     const users = await principalService.findAll();
-    ctx.response.body = hal.collection(users);
+    const stats = await principalService.getPrincipleStats();
+    ctx.response.body = hal.collection(users, stats);
```
evert, line 20: We want the stats on the home `/` not `/user` =)

evert (changes requested):

evert (approved):

PR #332, opened 2021-06-03
A new systems for settings.

evert:
I've been wanting to revamp settings for a bit. There's currently a few
different ways to provide settings, including environment variables and
via the database.

There is no central place where it's clear which setting exists, and
there's no good enough rules around where settings are set from
(environment or db).

Furthermore, if a setting is not set, we usually have a default. What
that default is, is determined by the 'caller', and also not centrally
which adds the risk that if different parts of the codebase use a
setting, there's nothing that enforces that that default is the same
everywhere.

Lastly, I want settings to bubble up in the interface so an admin can
easily see what settings are active (and where those settings came
from).

My intention with this PR is to solve all these problems, and so far
it's backwards compatible. This opens the door to some cool features,
and hopefully makes everything a little clearer.

### DIFF
--- a/src/server-settings.ts
+++ b/src/server-settings.ts
@@ -48,19 +48,13 @@ const settingsRules: SettingsRules = {
   'smtp.url' : {
     description: 'The url to the SMTP server. See the node-mailer documentation for possible values',
     env: 'SMTP_URL',
-
-    // There's no reason this can't be in the DB, but code using this setting just needs to
-    // be changed to use the settings api.
-    fromDb: false,
+    fromDb: true,
     default: null,
   },
   'smtp.emailFrom' : {
     description: 'The "from" address that should be used for all outgoing emails',
     env: 'SMTP_EMAIL_FROM',
-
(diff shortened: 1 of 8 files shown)


src/server-settings.ts
```diff
@@ -1,52 +1,249 @@
...
+
+    let msg = `A value for the setting "${setting}" must be provided for this feature to work. `;
+    if (info.fromDb && info.env) {
+      msg+=`You should either set the "${setting}" setting in the database, or provide a value with the "${info.env}" environment variable.`;
+    } else if (info.fromDb) {
+      msg+=`You should set the "${setting}" setting in the database.`;
+    } else if (info.env) {
+      msg+=`You should provide a value with the "${info.env}" environment variable.`;
+    } else {
+      msg+'There is literally no way to actually do this, and this is a bug. DM me for a prize.';
```
BeckyPollard, line 174: Nice

BeckyPollard (approved):
Resolve the failed test, then it looks good to go.

PR #333, opened 2021-06-08
sort group members alphabetical

juhangsin:


src/group/formats/hal.ts
```diff
@@ -45,6 +45,12 @@ export function collection(groups: Group[]): HalResource {
  */
 export function item(group: Group, privileges: PrivilegeMap, isAdmin: boolean, groups: Group[], members: Principal[]): HalResource {
 
+  const allMember = members.sort(function(a, b){
```
juhangsin, line 48: maybe this could be a until function, since i can see it being used in multiple places?

PR #342, opened 2021-06-28
user search by href will redirect to the user

juhangsin:
#296 searching user by href will redirect to the actual user.

### DIFF
--- a/src/user/controller/by-href.ts
+++ b/src/user/controller/by-href.ts
@@ -1,12 +1,6 @@
 import Controller from '@curveball/controller';
 import { Context } from '@curveball/core';
-import * as privilegeService from '../../privilege/service';
-import * as userHal from '../formats/hal';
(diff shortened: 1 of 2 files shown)


src/user/controller/by-href.ts
```diff
@@ -40,6 +40,7 @@ class UserByHrefController extends Controller {
           isAdmin,
           await groupService.findGroupsForPrincipal(principal),
         );
+        ctx.redirect(303, principal.href);
```
evert, line 43: With this addition, we can skip the rendering part (ctx.response.body doesn't need to be set).

Also, this change should probably just be applied to all 3 principal types... With that, I think the `switch` statement can actually be removed?

evert (approved):
Best PR ever?

juhangsin:
> Best PR ever?
:100:

Issue #295, opened 2021-04-06
Create users API and 'edit user' form don't check if `identity` is a valid URI.


evert:
A identity such as `foo@example` should be rejected.

PR #348, opened 2021-09-28
throw error for invalid user uri

juhangsin:
https://github.com/curveball/a12n-server/issues/295

this handles when `mailto:` isn't included for type user during POST for `create-user` & PUT for '/user/:userId/edit` process.

### DIFF
--- a/src/create-user/controller.ts
+++ b/src/create-user/controller.ts
@@ -28,6 +28,9 @@ class CreateUserController extends Controller {
     const type = ctx.request.body.type;
 
     if (!identity || !identity.includes(':') || identity.includes(' ')) {
+      if (type === "user" && !principalService.isIdentityValid(identity, type)) {
+        throw new UnprocessableEntity('Identity must include mailto:');
+      }
       throw new UnprocessableEntity('Identity must exist and must be a url');
     }
     if (nickname.length < 1) {
--- a/src/principal/service.ts
+++ b/src/principal/service.ts
@@ -1,4 +1,4 @@
-import { NotFound } from '@curveball/http-errors';
+import { NotFound, UnprocessableEntity } from '@curveball/http-errors';
 import database from '../database';
 import { Principal, NewPrincipal, PrincipalType, User, Group, App, PrincipalStats } from './types';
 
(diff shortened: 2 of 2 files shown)


evert:
Validating this is a good idea, but! We actually want to support all kinds of URIs, and not just `mailto:`. For example, a user might sign up with a `tel:` instead, or their blog (using `http`).

Maybe this can be changed to strictly validate that the identity is an absolute URI?

juhangsin:
new PR this should check if it starts with `mailto: tel: https:// http://`

src/principal/service.ts
```diff
@@ -250,6 +255,15 @@ export function recordToModel(user: PrincipalRecord): Principal {
 
 }
 
+export function isIdentityValid(identity: string, type: 'user'): boolean {
+
+  const regex = /^(?:[A-Za-z]+:)?/;
+  if (type === 'user' && regex.test(identity)) {
+    return true;
+  }
+  return false;
```
evert, line 264: This can be shortened to

```suggestion
  return regex.test(identity);
```

src/create-user/controller.ts
```diff
@@ -27,6 +27,10 @@ class CreateUserController extends Controller {
     const nickname = ctx.request.body.nickname;
     const type = ctx.request.body.type;
 
+    if (type === 'user' && !principalService.isIdentityValid(identity, type)) {
+      throw new UnprocessableEntity('Identity must include mailto:');
```
evert, line 31: Error is no longer correct. This should probably be something like `Identity must be a valid URI`.

One more note here, we started actually adopting some json-schema in a12nserver, so it's also possible to use that here and just fully do validation with json schema, but this is fine too.

src/create-user/controller.ts
```diff
@@ -27,6 +27,10 @@ class CreateUserController extends Controller {
...
+    }
+
     if (!identity || !identity.includes(':') || identity.includes(' ')) {
```
evert, line 34: This `if` statement seems to at least partially duplicate the new check you are doing. We should probably validate the identity just once.

evert: I do like the idea of having the check outside the controller and in the service, so it's reuable.

src/principal/service.ts
juhangsin: :exploding_head:

PR #356, opened 2021-12-30
Adding a broken attempt at injecting knex with types into a12n

mihok:
### Changes

Im attempting to include knex as the database client to enable PostgreSQL in addition to MySQL, and in the future enable all the database clients knex supports.

Currently trying to keep up with types, and providing a minimal attempt at not having to create any breaking changes.

I'm currently fighting with the `params` parameter for the `.query` function (which now just pipes directly into `knex.raw`. Their function signatures match, and in theory the types they use should match as well but I'm still having issues. (reference to all the knex types: https://github.com/knex/knex/blob/master/types/index.d.ts)


src/app.ts
```diff
@@ -27,8 +26,6 @@ if (!process.env.PUBLIC_URI) {
 (async () => {
 
   console.log('Connecting to database');
-  await checkPatches();
```
evert, line 30: I think we need to auto-run knex migrations on start. I believe knex ensures that if multiple a12nserver start at the same time, only 1 will run the migration (using table locks).

src/database.ts
```diff
@@ -1,88 +1,77 @@
...
-  let settings: any = {};
-
-  // We are running in a local environment.
-  settings = {
-    user: process.env.MYSQL_USER,
-    password: process.env.MYSQL_PASSWORD,
-    database: process.env.MYSQL_DATABASE,
-  };
+  let connection: any = {};
+  let client: string = '';
```
evert, line 35:
```suggestion
  let connection;
  let client;
```

I assume if you don't initialize these, typescript will complain. That's a good thing, because I think you just need an `else` statement with an explicit error.

src/database.ts
```diff
@@ -1,88 +1,77 @@
...
 }
 
 export default {
-  query,
   getConnection,
-  checkPatches,
+  query,
 };
 
 async function getSettings() {
```
evert, line 32: Could this return something like `Knex.Config` or `Knex.Settings` (I assume a type exists)

src/database.ts
```diff
@@ -1,88 +1,77 @@
...
+      connection.socketPath = '/cloudsql/' + process.env.MYSQL_INSTANCE_CONNECTION_NAME;
+    } else {
+      delete connection.host;
+      delete connection.port;
+    }
   }
-  return settings;
 
-}
+  // We are running in a local environment.
```
evert, line 65: I don't understand this comment =)

src/mfa/totp/service.ts
```diff
@@ -17,17 +17,19 @@ export function generateSecret(): string {
...
     user_id: totpDevice.user.id,
     secret: totpDevice.secret,
   };
 
-  const result = await database.query(query, [newTotpDeviceRecord]);
+  const connection = await database.getConnection();
+
+  const result = await connection<UserTotpDeviceRow>('user_totp')
+    .insert(newTotpDeviceRecord, 'id')
+    .returning('id')
```
evert, line 29: I'm fairly certain `.returning()`  breaks in MySQL. I recently ran into this myself. Knex (for me) throws a warning that it's not supported in MySQL.

evert: Also this might be missing a value for the `created` field

src/mfa/webauthn/service.ts
```diff
@@ -15,31 +15,34 @@ type UserWebAuthnRow = {
...
       public_key: webAuthNDevice.publicKey.toString('base64'),
       counter: webAuthNDevice.counter
     };
 
-    const result = await database.query(query, [newWebAuthnRecord]);
+    const connection = await database.getConnection();
+
+    const result = await connection<UserWebAuthnRow>('user_webauthn')
+      .insert(newWebAuthnRecord, 'id')
+      .returning('id');
```
evert, line 29: * [ ] created missing?
* [ ] returning might not work

src/oauth2-client/service.ts
```diff
@@ -118,26 +118,27 @@ export async function getOAuth2ClientFromBody(ctx: Context<any>): Promise<OAuth2
...
+  const connection = await db.getConnection();
+  const result = await connection<OAuth2ClientRecord>('oauth2_clients')
+    .insert(params, 'id')
+    .returning('id');
+
+  const newClient = await connection<OAuth2ClientRecord, OAuth2ClientRecord>('oauth2_client')
+    .select({ id: result[0] })
+    .returning('*');
+
+  const realClient = await mapToOauth2Client(newClient[0], client.app, client.allowedGrantTypes);
```
evert, line 137: Doesn't need async/await

src/oauth2-client/service.ts
```diff
@@ -152,3 +153,16 @@ export async function validateSecret(oauth2Client: OAuth2Client, secret: string)
...
 
+async function mapToOauth2Client(data: OAuth2ClientRecord, app: App, allowedGrantTypes: GrantType[]): Promise<OAuth2Client> {
+  return {
+    id: data.id,
+    clientId: data.client_id,
+    clientSecret: data.client_secret.toString(),
+    app: {
+      ...app,
+      id: data.user_id,
+    },
```
evert, line 164: Would this work?

```suggestion
    app,
```

src/database.ts
evert: I don't think this function needs to be `async`. Maybe this was wrong before.

src/app.ts
evert: I think we need to migrate the migrations to knex.

I don't think this means a full-rewrite of all migrations to knex migrations, but I do think we need a knex migration that:

1. Checks the patch level
2. If it's below the latest patch (from the old patch system), either:
  * complain and die and tell users to manually apply these patches
  * Or: automatically run the patches (with `.raw` ?)

This is to support an upgrade path from an earlier a12nserver.

evert (changes requested):
What would be really nice is if:

* We could get some github actions test that sets up a Postgres or MySQL and see if it can start a12nserver, maybe a rudimentary test?
* Types for knex records, see: https://github.com/evert/tt/blob/main/src/knex.ts

src/mfa/totp/service.ts
mihok: Yeah, so line 28's additional parameter is what allows MySQL to return as expected, so to support both databases you have to do line 28 and line 29, `.returning('id')` wont return the id in MySQL on its own.

src/database.ts
evert: This is not 100% resolved, because initializing these variables is still a bad idea.

src/app.ts
mihok: Hmmm, I think I might go with the complain and die approach for now...

evert: Yeah i think that's fine. This will at least prevent people from starting the server with a broken database.

src/database.ts
mihok: Ohhh I didnt fully understand your initial comment before! I'll fix

Makefile
```diff
@@ -20,15 +20,24 @@ test:
...
 lint-fix: fix
 
+migrate:
+	npx knex migrate:latest
+
+migrate-rollback:
+	npx knex migrate:rollback
+
+migration:
+	npx knex migrate:make $(name) -x ts
```
evert, line 37: :+1:

Makefile
```diff
@@ -43,5 +52,7 @@ dist/build: $(SOURCE_FILES)
 	touch dist/build
 
 inspect: build
-	node --inspect dist/app.js
+	node --inspect dist/src/app.js
```
evert, line 55: I hope we can revert this change. I ran into the same issue and it was because `knexfile.ts` is in the root.

Can we move `knexfile.ts` into `src/` ? The `npx knex` command has a `--knexfile` option to point to the right place.

I also did this here:

https://github.com/badgateway/tt/blob/main/package.json#L14

knexfile.ts
```diff
@@ -0,0 +1,12 @@
+// Update with your config settings.
```
evert, line 1:
```suggestion
```

migrations/20220105195635_init.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (result[0].id < MAGIC_CHANGELOG_NUMBER) {
+      throw new Error('Migrations must have reached `changelog` version 42 before applying Knex migrations, please apply those manually and then run `make migrate`.');
+    }
+
+  } catch (err: any) {
+    // Color the error red.
+    // eslint-disable-next-line no-console
+    console.error('\x1b[31m', err, '\x1b[37m');
+
+    process.exit(1);
```
evert, line 27: Instead of `exit,` you can also just `throw err` or just never catch this error at all (Knex should handle errors)

src/app.ts
```diff
@@ -27,8 +26,7 @@ if (!process.env.PUBLIC_URI) {
...
-
+  console.log('Running Migrations');
```
evert, line 29: Is this a lie?

evert (commented):
Just a few more requests.

Great PR. It doesn't just add major features (postgres support, better migrations) but also improves most of the code it touches

src/app.ts
mihok: Lol it happens when the pool connection is made but seemed weird with the logs going:

1. Connecting to database
2. Reading settings
3. Running Migrations

So I put it with the others, which are technically a lie too lol, I can move the console logs into the right places, I think thats the most correct

migrations/20220105195635_init.ts
mihok: Yeah let me remove the try catch all together

Makefile
mihok: Anytime that we add another entry in tsconfig.js that is _outside_ of the `src/` directory it will distrupt how the `dist/` folder willl look. So if we want to revert the `dist/src/app.js` changes we would have to be okay with both the `knexfile.ts` and `migrations/` residing in `src/`, is that preferred?

evert: I think so, yes!

src/app.ts
evert: Yeah can we? I imagine knex output will be interspersed with these too

migrations/20220105195635_init.ts
```diff
@@ -0,0 +1,36 @@
+import { Knex } from 'knex';
+
+// Funny how this number ended up being the last changelog migration...
+const MAGIC_CHANGELOG_NUMBER = 42;
```
trasherdk, line 4: Well.. It is the `Answer to the Ultimate Question of Life, the Universe, and Everything,` :rofl:

Makefile
```diff
@@ -27,6 +27,15 @@ fix:
...
 lint-fix: fix
 
+migrate: build
+	npx knex migrate:latest --knexfile './dist/knexfile.ts'
+
+migrate-rollback: build
+	npx knex migrate:rollback --knexfile './dist/knexfile.ts'
+
+migration: build
+	npx knex migrate:make $(name) -x ts --knexfile './dist/knexfile.ts'
```
evert, line 37:
```suggestion
```

Lets just remove this? It's less useful i think, and the name might confuse people

package.json
```diff
@@ -7,8 +7,8 @@
     "prepublishOnly": "make build",
     "test": "make lint test",
     "tsc": "tsc",
-    "start": "node ./dist/app.js",
-    "start:watch": "tsc-watch --onSuccess 'node dist/app.js'"
+    "start": "node ./dist/src/app.js",
+    "start:watch": "tsc-watch --onSuccess 'node dist/src/app.js'"
```
evert, line 11: Possibly still wrong? We need to get that `src` out of there again

src/group/service.ts
```diff
@@ -32,20 +32,17 @@ export async function addMember(group: Group, user: Principal): Promise<void> {
...
     }
-    await connection.commit();
+    await transaction.commit();
   } catch (err) {
-    await connection.rollback();
+    await transaction.rollback();
     throw err;
-  } finally {
-    connection.release();
   }
```
evert, line 45: This is a red flag to me. Most database adapters will give the user a dedicated 'connection' when using `getConnection()`, until it's released back to the pool. Is this not the case with knex?

evert (changes requested):
The biggest thing i can think of right now is .. do the installation instructions still make sense with the new knex migrations?

Makefile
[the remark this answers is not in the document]
mihok: So its okay for folks to generate their own migration files? `migrate` runs the migrations, this makes a new migration file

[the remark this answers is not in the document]
evert: Renamed this to `create-migration`

src/migrations/001_initial.ts
```diff
@@ -0,0 +1,60 @@
...
+    table.integer('timestamp').unsigned();
+
+  });
+
+  await knex('changelog').insert({
+    id: 1,
+    timestamp: Math.floor(Date.now()/1000)
+  });
+
+  await knex.schema.createTable('users', table => {
```
evert, line 22: @mihok This is an example of a migration that I edited to work with postgres. I only did path 001 to 004 so far.

src/migrations/002_oauth2.ts
```diff
@@ -0,0 +1,39 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+
+  const result = await knex('changelog').select('id').where({id: 2});
+
+  if (result.length) {
+    // Migration has been applied using the old patch system
+    return;
+  }
```
evert, line 10: @mihok  These lines ensure that we don't re-run migrations that were run using the old system. This also means that if someone uses your 'postgres_schema/042-transfer-from-mysql.sql` file, these files also safely get skipped.

src/migrations/005_oauth2-redirect-uri.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (result.length) {
+    // Migration has been applied using the old patch system
+    return;
+  }
+  await knex('changelog').insert({
+    id: 5,
+    timestamp: Math.floor(Date.now()/1000)
+  });
+
+  await knex.raw(`CREATE TABLE oauth2_redirect_uris (
```
evert, line 16: @mihok most migrations look like this, and only work with mysql. I wrote a script to generate these files automatically.

.github/workflows/test.yml
```diff
@@ -11,20 +11,71 @@ on:
...
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm ci
-    - run: npm run build --if-present
-    - run: npm test
+    - run: make test
+  postres:
```
mihok, line 52: Improperly named I think? Given the mysql image

src/migrations/003_oauth-access-token.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+
+  const result = await knex('changelog').select('id').where({id: 3});
+
+  if (result.length) {
+    // Migration has been applied using the old patch system
+    return;
+  }
+  await knex('changelog').insert({
```
mihok, line 11: We've done away with transactions here, is this intentional?

src/migrations/005_oauth2-redirect-uri.ts
mihok: We'll have to go back through them as I think some of PostgreSQLs migration syntax is slightly different (different types) but this is a good first step having these files generated.

src/migrations/011_user-type.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  await knex('changelog').insert({
+    id: 11,
+    timestamp: Math.floor(Date.now()/1000)
+  });
+
+  await knex.raw(`ALTER TABLE oauth2_clients
+  DROP name`);
+  await knex.raw(`ALTER TABLE users
+  ADD type TINYINT UNSIGNED NOT NULL DEFAULT '1' COMMENT '1 = user, 2 = app'`);
```
mihok, line 19: For example @evert this wont run in PostgreSQL because there is no `tinyint` type

mihok (commented):
These changes look good @evert , its a good step to get the code in there but still rely on MySQL for now. I think once I have some freetime I'll go through the migrations and update them to work for both SQL dbs

.github/workflows/test.yml
evert: Yes! Will fix

src/migrations/003_oauth-access-token.ts
evert: Knex wraps these migrations in something that basically works the same. The transactions in the raw sql files were helpful for certain edge cases (specifically until the fist CREATE/ALTER table that has a syntax error, but not after). I believe knex works the same.

src/migrations/011_user-type.ts
evert: Yes, but if you run the file you generated (from posgres-schema/042-migration-from-mysql.sql) then all these lines should be skipped. It's a workaround until all these migrations are rewritten but it should work

mihok: Right good point.

evert (approved):

PR #358, opened 2022-01-04
Features/logo

ikbensiep:
Adds a logo to the `/login` page. 
Logo is a link in the `server_settings` table, or if absent there, the Curveball logo (hardcoded to curveballjs.org). 

The PR is still incomplete, _I think_ the logo should be added to all pages, but up for suggestions. 
I tried to looking into that but I'm not sure how to add the prop to the main layout.hbs template, @evert do you have some suggestions ?

Also note how the parameter is typed in `formats/html.ts`, not sure if that\s the best way to do it, the other option was to have `await getSetting('logo_url') || ''` in the controller, which seems ugly.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -112,3 +112,27 @@ The SMTP_URL takes any format that that [Nodemailer](https://nodemailer.com/smtp
 
 
 [oauth2-jwt]: https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt-12
+
+
+Server Settings & Defaults
+--------------------------
+
+| Name                      | Type | Default value | Description |
+|:--------------------------|------|--------------:|-------------|
+| cors.allowOrigin          | Integer | NULL          | List of allowed origins that may directly talk to the server. This should only ever be 1st party, trusted domains. By default CORS is not enabled     
+| jwt.privateKey            | String  | NULL | The RSA private key to sign JWT access tokens. Usually this value has the contents of a .pem file. If not set, JWT will be disabled
+| login.defaultRedirect     | String  | `/`         | The url that the user will be redirected to after the log in to a12nserver, and no other redirect_uri is provided by the application. It's a good idea to set this to your application URL. 
+| logo_url                  | String  | NULL | The application logo to display on the a12nserver pages. If no logo url is provided, the application will show the Curveball logo by default. |
+| oauth2.code.expiry        | Integer | 600 (10 minutes) | The expiry time (in seconds) for the \'code\' from the oauth2 authorization code grant type |
+| oauth2.accessToken.expiry | Integer | 600 (10 minutes) | The expiry time (in seconds) for OAuth2 access token. |
+| oauth2.refreshToken.expiry| Integer | 3600 * 6 (6 hours) | The expiry time (in seconds) for OAuth2 refresh tokens. |
+| registration.enabled      | Boolean | true    | Allow users to register new accounts. By default new accounts will be disabled and have no permissions.|
+| registration.mf.enabled   | Boolean | true | Allow users to register new accounts. By default new accounts will be disabled and have no permissions.
+| smtp.url                  | String  | NULL  | The url to the SMTP server. See the node-mailer documentation for possible values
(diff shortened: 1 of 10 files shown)


src/login/controller/login.ts
```diff
@@ -31,6 +31,7 @@ class LoginController extends Controller {
         continue: ctx.query.continue,
       },
       await getSetting('registration.enabled'),
+      await getSetting('logo_url')
```
evert, line 34: This doesn't need `await` (and neither does the call right above it!)

src/server-settings.ts
```diff
@@ -35,7 +36,11 @@ type SettingsRules = {
 }
 
 const settingsRules: SettingsRules = {
-
+  'logo_url' : {
+    description: 'The application logo to display on the a12nserver pages',
+    fromDb: true,
+    default: '' //curveball logo
```
evert, line 42: The curveball logo is already part of this codebase. After logging in, you should see it in the top-left. So this value should be something like: `/_hal-browser/assets/logo.svg` (I don't know the actual path).

templates/login.hbs
```diff
@@ -1,4 +1,17 @@
...
+    <!-- show app logo -->
+    <fieldset><img src={{logoUrl}} style="max-width: 100%"/></fieldset>
+    {{#else}}
+    
+    <!-- show curveball logo -->
+    <fieldset style="background: #1e88e5;">
+      <img src="https://curveballjs.org/assets/images/logo.svg" style="max-width: 100%"/>
+    </fieldset>
+    {{/else}}
+  {{/if}}
```
evert, line 13: * `fieldset` is a bit strange, no?
* Doesn't need the `else` statement, if the 'default logo' is provided by the settings instead.
* Could the logo instead be added to the 'parent' template, so it shows up on registration, login, lost password, etc?

pschwyter:
Can you also include some screenshots so we can see what's changed?

pschwyter:
Can you also update the README to mention this feature? Probably just a section in the `getting-started.md` file?

src/server-settings.ts
```diff
@@ -21,6 +21,7 @@ type Settings = {
   'webauthn.expectedOrigin': string | null;
   'webauthn.relyingPartyId': string | null;
   'webauthn.serviceName': string;
+  'logo_url': string | null;
```
evert, line 24: You can remove the `| null` part everywhere., just provide a good default.

src/templates.ts
```diff
@@ -16,10 +17,12 @@ export function render(name: string, params?: Params): string {
...
   Object.assign(newParams, params, {
     appName: 'Auth API',
+    logoUrl: getSetting('logo_url')
   });
+
   return layoutTemplate({
-    ...params,
-    body: pageTemplate(params)
+    ...newParams,
+    body: pageTemplate(newParams)
```
ikbensiep, line 25: In the old code, nothing was done with `newParams`, also resulting in the page's `<title>` attribute not being populated correctly.

evert (approved):

pschwyter (approved):
I'd suggest updating the README to document this feature. But if evert approves, I approve.

ikbensiep:
@evert @pschwyter I updated the getting started doc.  
This is not my expertise; perhaps it should be moved and/or have a paragraph added? Please review, cheers!

docs/getting-started.md
```diff
@@ -112,3 +112,27 @@ The SMTP_URL takes any format that that [Nodemailer](https://nodemailer.com/smtp
 
 
 [oauth2-jwt]: https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt-12
+
+
+Server Settings & Defaults
```
evert, line 117: Lets put this in its own document.

Some suggested flavor text:

The following table lists all configurable options for a12nserver. These can be changed in the `server_settings` table in your database.

The value for each field is stored as a JSON string. This means that in order to store `null`, you need the literal string `null`. To store a string, it needs to be surrounded by double-quotes. For example, a legal value for the `smpt.emailFrom` field might be `"info@example.org"`

evert:
Just needs the new logo and the new browser package and this is good to go.

PR #363, opened 2022-01-11
Revert "Revert "Features/logo""

ikbensiep:
Reverts curveball/a12n-server#360 

Reinstates previous logo / server settings docs updates 

Updates curveball logo to include outline, drop shadows

login page preview:
image

### DIFF
--- a/src/server-settings.ts
+++ b/src/server-settings.ts
@@ -39,7 +39,7 @@ const settingsRules: SettingsRules = {
   'logo_url' : {
     description: 'The application logo to display on the a12nserver pages',
     fromDb: true,
-    default: '/_hal-browser/assets/curveball-logo-wordmark.svg'
+    default: '/_hal-browser/assets/curveball-logo.svg'
   },
   'login.defaultRedirect': {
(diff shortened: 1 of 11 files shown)


evert:
One thing that's probably missing from this PR, is that it needs the updated `@curveball/browser` dependency. So get that released, and then we can do this one =)

src/server-settings.ts
```diff
@@ -35,7 +36,11 @@ type SettingsRules = {
 }
 
 const settingsRules: SettingsRules = {
-
+  'logo_url' : {
+    description: 'The application logo to display on the a12nserver pages',
+    fromDb: true,
+    default: '/_hal-browser/assets/curveball-logo.svg'
```
evert, line 42: When I looked at this last night, I *think* this is not the right path for what was released in the curveball browser, so I would unlink your local env and try this with the latest browser package.

evert: My bad, was looking at an old version of the browser package.

evert (approved):

PR #365, opened 2022-03-21
Updated Permissions

syedfkabir:
Edited the instructions so that ALTER, CREATE, and DROP privileges 
are granted to allow the following script to run successfully: 
cat mysql-schema/*.sql | mysql -u a12nserver -p -h hostname a12nserver


pschwyter:
Thanks for updating these docs. Could you add a quick description for the change?

pschwyter (approved):
thanks!

PR #369, opened 2022-04-01
All migrations now run on Postgres

evert:


src/migrations/021_access-token-fix.ts
```diff
@@ -13,8 +13,17 @@ export async function up(knex: Knex): Promise<void> {
     timestamp: Math.floor(Date.now()/1000)
   });
 
-  await knex.raw(`ALTER TABLE oauth2_tokens 
-  CHANGE access_token access_token VARCHAR(2000) CHARACTER SET ascii`);
+  switch(knex().client.driverName) {
+    case 'mysql' :
```
mihok, line 17: Does the tsconfig allow for fallthrough cases ;)

Also out of curiosity, why the split here?

mihok (approved):
Looks great!

src/migrations/021_access-token-fix.ts
evert: On MySQL it errors because the index size is too big. It's in part because we use utf8mb4, which causes the index size to be 4 * 2000 bytes = 8000 bytes. Setting the character set for the field to `ascii` makes it top out at 2000 bytes.

evert: And yeah fallthtrough cases are forbidden except if they're empty!

PR #373, opened 2022-04-03
user_totp: Unknown column 'user_id' in 'field list'

### LINKED ISSUE #372: user_totp: Unknown column 'user_id' in 'field list'
Taking this thing for a first spin, and run into this:
```
=> (15) POST /register/mfa/totp
.returning() is not supported by mysql and will not have any effect.
Error: insert into `user_totp` (`created`, `secret`, `user_id`) values (1648938461, 'GJKVSODRENTVWFY5', 1) - Unknown column 'user_id' in 'field list'
```
Makes sense when `src/migrations/001_initial.ts` says:
```js
  await knex.schema.createTable('user_totp', table => {

    table.integer('user_totp').unsigned().notNullable().primary();
    table.string('secret', 50);
    table.tinyint('failures').unsigned().notNullable().defaultTo(0);

  });
```
After migrations it loks like this:
```sql
CREATE TABLE a12ntest.user_totp (
  user_totp INT(10) UNSIGNED NOT NULL,
  secret VARCHAR(50) DEFAULT NULL,
  failures TINYINT(3) UNSIGNED NOT NULL DEFAULT 0,
  created INT(10) UNSIGNED NOT NULL DEFAULT 0,
  PRIMARY KEY (user_totp)
)
ENGINE = INNODB,
CHARACTER SET utf8,
COLLATE utf8_unicode_ci;
```
And then I got this `src/user/service.ts`:
```js
export async function validateTotp(user: User, token: string): Promise<boolean> {

  const query = 'SELECT secret FROM user_totp WHERE user_id = ?';
  const result = await database.query(query, [user.id]);

  if (!result.length) {
    // Not set up
    return false;
  }
  const secret = result[0].secret;

  return otplib.authenticator.check(token, secret);

}
```

trasherdk:
fix: user_totp - change user_totp.user_totp to user_totp.user_id
add flush-mysql.sh (drop all tables)

closes #372

### DIFF
--- a/src/migrations/001_initial.ts
+++ b/src/migrations/001_initial.ts
@@ -1,6 +1,6 @@
 import { Knex } from 'knex';
 
-export async function up(knex: Knex): Promise<void> {
+export async function up (knex: Knex): Promise<void> {
 
   if (await knex.schema.hasTable('changelog')) {
     // This migration has been applied
@@ -16,7 +16,7 @@ export async function up(knex: Knex): Promise<void> {
 
   await knex('changelog').insert({
     id: 1,
-    timestamp: Math.floor(Date.now()/1000)
+    timestamp: Math.floor(Date.now() / 1000)
   });
 
   await knex.schema.createTable('users', table => {
@@ -45,7 +45,7 @@ export async function up(knex: Knex): Promise<void> {
 
   await knex.schema.createTable('user_totp', table => {
 
-    table.integer('user_totp').unsigned().notNullable().primary();
+    table.integer('user_id').unsigned().notNullable().primary();
     table.string('secret', 50);
     table.tinyint('failures').unsigned().notNullable().defaultTo(0);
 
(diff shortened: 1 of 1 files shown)


evert:
Thanks for the PR! the change looks good, but could you perhaps remove `flush-mysql.sh` ?

trasherdk:
Yes, I can remove the script. I thought it's a handy tool during testing.

Is adding script to .gitignore, and move flush-mysql.sh in there, okay?

evert:
> Yes, I can remove the script. I thought it's a handy tool during testing.
> 
> Is adding script to .gitignore, and move flush-mysql.sh in there, okay?

One idea might be to just add it to a `~/.bin` and add that to your `PATH` variable? Alternatively you can set it up as a `alias` in your shell, so it's available when you need it but we don't need to put something in `~/.gitignore` just for you.

trasherdk:
No problem. It's gone.

evert (approved):

PR #377, opened 2022-04-16
Some changes to settings

evert:
This PR

* Introduces an admin-only 'settings' page which lets admins see what settings have been applied, and in the future maybe even change settings.
* Changes some of the environment variables for database setup. Now there's a generic `DB_HOST` instead of a `PG_HOST` and `MYSQL_HOST`, but the other 3 are still supported for backwards compatibility.
* Automatically generates a local sqlite3 database in a file instead of memory, so developers can get started very fast. Sqlite3 still needs work before it fully works though.
* Fixes a bug with 404 pages resulting a totally blank response.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -54,8 +54,6 @@ export async function query<T = any>(query: string, params: Knex.ValueDict | Kne
 
 }
 
-
-
 export function getSettings(): Knex.Config {
 
   let connection: Knex.MySql2ConnectionConfig | Knex.PgConnectionConfig | Knex.Sqlite3ConnectionConfig;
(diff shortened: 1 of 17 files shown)


src/database.ts
```diff
@@ -17,14 +19,53 @@ export async function init() {
...
+
+type RawResult<T> = RawMySQLResult<T> | RawPostgreSQLResult<T> | RawSqlite3Result<T>;
+
+
+
+/**
+ * A shortcut for easily executing select queries.
+ *
+ * Use of this should be phased out, but it helped with the migration from
+ * befre knex.
```
mhum, line 37:
```suggestion
 * before knex.
```

mhum (approved):

PR #385, opened 2022-05-03
Fix knex module augmentation not working

lorefnon:
Resolves https://github.com/knex/knex/issues/5156


evert (approved):
Works! I swear I thought I tried this. Thank you!!!

PR #392, opened 2022-05-13
Fixed reset password

pschwyter:
Found the issue, we weren't properly accessing db results


evert (approved):
Glad these are surfacing! So much code has changed

PR #394, opened 2022-05-19
Fix Basic Auth header parsing

pschwyter:
The regex is very permissive, let me know if it should be more restricted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ migrate-rollback:
 	npx knex migrate:rollback --knexfile './dist/knexfile.js'
 
 start-dev:
-	npx tsc-watch --onSuccess 'node dist/app.js'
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ migrate-rollback:
 	npx knex migrate:rollback --knexfile './dist/knexfile.js'
 
 start-dev:
-	npx tsc-watch --onSuccess 'node dist/app.js'
+	npx tsc-watch --onSuccess 'node --inspect=9339 dist/app.js'
```
evert, line 37: Are there any drawbacks to this?

pschwyter: nope. only benefits

pschwyter: the only drawback I can think of is if someone is running multiple node apps and has already assigned a debugger to port `9339` it will complain (default is `9229`).

evert (approved):

PR #404, opened 2022-06-24
requirePkce setting was flipped

evert:


src/oauth2-client/service.ts
```diff
@@ -148,7 +148,7 @@ export async function edit(client: OAuth2Client, redirectUris: string[]): Promis
 
   const params: Partial<OAuth2ClientRecord> = {
     allowed_grant_types: client.allowedGrantTypes.join(' '),
-    require_pkce: client.requirePkce?0:1,
+    require_pkce: client.requirePkce?1:0,
```
pschwyter, line 151: Here, you dropped these: `  `

pschwyter: not even a chuckle

evert: i just chuckled

evert: SORRY

Issue #399, opened 2022-06-19
Don't throw 500 on conflicts when creating a new oauth2 client 


evert:
If you're creating a new OAuth2 client, and you're using a `client_id` that's already in the system, a12nserver will emit a 500 error due to a violation of the unique constraint.

It would be nice if instead we threw a more user-friendly error. This should probably be a `409 conflict`.

To create a new OAuth2 client:

1. Go to 'Apps'
2. Create new App.
3. Find the new App resource you created, there should be an option to see the 'oauth2 clients'.
4. From there, there should be a navigation item to create a new client.

PR #411, opened 2022-07-24
Added 409 Conflict for OauthClient Add #399

AminDhouib:
Fix for this issue: https://github.com/curveball/a12n-server/issues/399

### DIFF
--- a/src/oauth2-client/controller/collection.ts
+++ b/src/oauth2-client/controller/collection.ts
@@ -2,13 +2,12 @@ import Controller from '@curveball/controller';
 import { Context } from '@curveball/core';
 import * as privilegeService from '../../privilege/service';
 import * as hal from '../formats/hal';
-import { Forbidden, UnprocessableEntity, Conflict } from '@curveball/http-errors';
+import { Forbidden, UnprocessableEntity } from '@curveball/http-errors';
 import { findByApp, create } from '../service';
 import * as principalService from '../../principal/service';
 import { GrantType, OAuth2Client } from '../types';
 import * as bcrypt from 'bcrypt';
 import { generatePublicId, generateSecretToken } from '../../crypto';
-import { UniqueViolationError } from 'db-errors';
 
 class ClientCollectionController extends Controller {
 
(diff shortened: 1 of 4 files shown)


src/oauth2-client/controller/collection.ts
```diff
@@ -73,6 +74,16 @@ class ClientCollectionController extends Controller {
       requirePkce: ctx.request.body.requirePkce ?? false,
     };
 
+    // Verify if client id already exists. Throw error if so.
+    try {
+      await findByClientId(clientId);
```
mhum, line 79: The safer way to do this is to catch and check the error returned by the DB when doing the insert. In this scenario, two requests coming in at nearly the same time could both pass this check but one still fail the DB-level constraint.

evert: I agree, there could be other problems than 'Not Found' that shouldn't be translated into a `Conflict`.

Also it might be a bit better if this check is moved to inside the service.

src/oauth2-client/controller/collection.ts
```diff
@@ -73,8 +73,18 @@ class ClientCollectionController extends Controller {
...
     };
 
-    const client = await create(newClient, redirectUris);
-    ctx.response.body = hal.newClientSuccess(client, redirectUris, clientSecret);
+    // If client id already exists in DB, throw error.
+    try {
+      const client = await create(newClient, redirectUris);
+      ctx.response.body = hal.newClientSuccess(client, redirectUris, clientSecret);
+    } catch (error: any) {
+      const err = wrapError(error);
```
evert, line 81: Awesome solution and great library.

A last thing I would change is to move this entire check to inside the service. The advantage is that if other code uses this `create` function, they just get nice errors.

Ideally the controller doesn't have to 'know' about databases. From a controllers perspective the service is a black box and the controller doesn't know if the service uses a database to store its information or not.

src/oauth2-client/controller/collection.ts
```diff
@@ -73,8 +74,17 @@ class ClientCollectionController extends Controller {
...
     };
 
-    const client = await create(newClient, redirectUris);
-    ctx.response.body = hal.newClientSuccess(client, redirectUris, clientSecret);
+    // If client id already exists in DB, throw error.
+    try {
+      const client = await create(newClient, redirectUris);
+      ctx.response.body = hal.newClientSuccess(client, redirectUris, clientSecret);
+    } catch (error: any) {
+      if (error instanceof UniqueViolationError) {
```
mhum, line 82: Better for the service to check for this error and throw `Conflict`. This is still a bit too much data access and DB concern being handled by the controller.

AminDhouib: Updated PR :)

mhum (approved):
Woot!

Issue #406, opened 2022-07-19
When adding a new users to a group, the screen goes white


evert:
Steps to reproduce:

1. Create a group
2. Add a member

Screen goes white but the operation does succeed.

I think at least part of why this is happening is that the form is handled using this tool:

https://github.com/badgateway/html-form-enhancer

When the form is submitted with this, the server responds with a `303 See Other` redirect back to itself. The intention for this was so the browser refreshes the screen, but it doesn't.

I suspect this may be a bit of a rabbit hole

PR #417, opened 2022-07-27
Fixed white screen when add/remov-ing mem from group

AminDhouib:
Fixed: https://github.com/curveball/a12n-server/issues/406

### DIFF
--- a/src/group/controller/item.ts
+++ b/src/group/controller/item.ts
@@ -97,7 +97,18 @@ class GroupController extends Controller {
     }
 
     if (ctx.request.accepts('text/html')) {
-      ctx.redirect(303, group.href);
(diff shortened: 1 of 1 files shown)


src/group/controller/item.ts
```diff
@@ -97,7 +97,18 @@ class GroupController extends Controller {
     }
 
     if (ctx.request.accepts('text/html')) {
-      ctx.redirect(303, group.href);
+      const group = await principalService.findByExternalId(ctx.params.id, 'group');
```
evert, line 100: You already have the data for `group` from earlier in the function

AminDhouib: Fixed :)

mhum (approved):

Issue #416, opened 2022-07-27
Make it easier to set up oauth2 clients for development environments.


evert:
It would be really nice if we can auto-generate a link that pre-fills some of the forms. I think the work for this should be done in several steps:

1. Add query parameters to the endpoint for creating new apps that pre-fills information for an app. For example `http://localhost:8531/app/new?name=A&url=B`.
2. A similar approach to the 'create new oauth2 client' form: `http://localhost:8531/app/[appid]/client/new`
3. As a third step, add all the query parameters from step 2, to the 'create new app' form from step 1. If some of these query parameters are set, the form from Step 1 should automatically redirect to the oauth2 form immediately after the app is created.

All of these should be done as separate PRs

PR #418, opened 2022-07-28
Issue #416: Step 1 Done

AminDhouib:
Fixing this issue (1st step): https://github.com/curveball/a12n-server/issues/416

### DIFF
--- a/src/app/controller/new.ts
+++ b/src/app/controller/new.ts
@@ -19,7 +19,7 @@ class CreateAppController extends Controller {
       throw new Forbidden('Only users with the "admin" privilege can create new users');
     }
     ctx.response.type = 'text/html';
-    ctx.response.body = createAppForm(ctx.query.msg, ctx.query.error, ctx.query);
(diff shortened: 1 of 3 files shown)


src/app/controller/new.ts
```diff
@@ -19,7 +19,7 @@ class CreateAppController extends Controller {
       throw new Forbidden('Only users with the "admin" privilege can create new users');
     }
     ctx.response.type = 'text/html';
-    ctx.response.body = createAppForm(ctx.query.msg, ctx.query.error);
+    ctx.response.body = createAppForm(ctx.query.msg, ctx.query.error, ctx.query);
```
mhum, line 22: Can you pass the `name` and `url` queries separately like the `msg` and `error` ones? Then, we can type those function parameters.

AminDhouib: Will do!

AminDhouib: Done :)

mhum (approved):

PR #437, opened 2022-10-03
Keep track of which apps requested which scopes for users

### LINKED ISSUE #435: Track which applications used which scopes for a user.

evert:
This enables us in the future to create the concept of a third-party apps. These apps will explicitly have to request the user if they are willing to give permission to certain scopes.

Right now all scopes are auto-granted, because all apps are considered 'first party' and trusted.

This is also a step towards OpenID Connect support.

Fixes #435

### DIFF
--- a/src/db-types.ts
+++ b/src/db-types.ts
@@ -1,3 +1,4 @@
+import 'knex';
 
 declare module 'knex/types/tables' {
 
@@ -12,6 +13,7 @@ interface Tables {
   privileges: PrivilegesRecord;
   reset_password_token: ResetPasswordTokenRecord;
   server_settings: ServerSettingsRecord;
+  user_app_permissions: UserAppPermissionsRecord;
   user_log: UserLogRecord;
   user_passwords: UserPasswordsRecord;
   user_privileges: UserPrivilegesRecord;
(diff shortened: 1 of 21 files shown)


src/db-types.ts
```diff
@@ -121,6 +123,40 @@ export type ServerSettingsRecord = {
   value: string | null;
 }
 
+export type UserAppPermissionsRecord = {
+  id: number;
+
+  /**
+   * Reference to the app / OAuth2 client
+   */
+  app_id: number;
```
pschwyter, line 132: Is this different from `client_id`?

src/user-app-permissions/controller/user-collection.ts
```diff
@@ -0,0 +1,28 @@
...
+import * as principalService from '../../principal/service';
+import * as userAppPermissionService from '../service';
+
+class UserAppPermissionsCollection extends Controller {
+
+  async get(ctx: Context) {
+
+    const principal = await principalService.findByExternalId(ctx.params.id, 'user');
+    if (ctx.auth.equals(principal) && !await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('You can only use this API for yourself yourself, or if you have \'admin\' privileges');
```
pschwyter, line 15:
```suggestion
      throw new Forbidden('You can only use this API for yourself, or if you have \'admin\' privileges');
```

src/user-app-permissions/controller/user-item.ts
```diff
@@ -0,0 +1,31 @@
...
+import * as principalService from '../../principal/service';
+import * as userAppPermissionService from '../service';
+
+class UserAppPermissionsItem extends Controller {
+
+  async get(ctx: Context) {
+
+    const user = await principalService.findByExternalId(ctx.params.id, 'user');
+    if (ctx.auth.equals(user) && !await privilegeService.hasPrivilege(ctx, 'admin')) {
+      throw new Forbidden('You can only use this API for yourself yourself, or if you have \'admin\' privileges');
```
pschwyter, line 15:
```suggestion
      throw new Forbidden('You can only use this API for yourself, or if you have \'admin\' privileges');
```

pschwyter (approved):
Just to make sure I understand: With this change apps will be able to explicitly request a list of scopes (as opposed to defaulting to *all* scopes), but this doesn't yet actually require permission from the user yet?

evert:
> Just to make sure I understand: With this change apps will be able to explicitly request a list of scopes (as opposed to defaulting to all scopes), but this doesn't yet actually require permission from the user yet?

Yeah this is correct. We currently only have 'privileged' applications. Auto-granting these scopes makes sense for first party-applications. But for third-party applications users would have to explicitly consent to these scopes.

This entire PR is really to set up the plumbing to make that possible in the future. Tracking scopes for privileged applications is not super useful by itself.

PR #438, opened 2022-10-10
Oauth2 JWT support (RFC 9068)

evert:


mhum (approved):
Do we actually support it now?

evert:
@mhum yup! JWT has worked for a while (a year?) but it needed a few tweaks to be compliant with RFC9068

PR #444, opened 2022-10-25
changed name to nickname

syedfkabir:
## Changes

we are looking for the query param for nickname, not name!

### consistent naming, nickname throughout a12n 

Screen Shot 2022-10-25 at 3 53 08 PM

### DIFF
--- a/src/app/controller/new.ts
+++ b/src/app/controller/new.ts
@@ -28,7 +28,7 @@ class CreateAppController extends Controller {
       csrfToken: await ctx.getCsrf(),
       msg: ctx.query.msg,
       error: ctx.query.error,
-      name: ctx.query.nickname,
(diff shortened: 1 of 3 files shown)


src/app/controller/new.ts
```diff
@@ -28,7 +28,7 @@ class CreateAppController extends Controller {
       csrfToken: await ctx.getCsrf(),
       msg: ctx.query.msg,
       error: ctx.query.error,
-      name: ctx.query.name,
+      name: ctx.query.nickname,
```
evert, line 31: Would you mind taking this 1 step further and also rename the property from `name` to `nickname` so it's consistent throughout the application? This likely also means updating the template

evert (approved):

PR #447, opened 2022-11-01
Bare minimum to support OpenID Connect

evert:
I believe this is absolute minimum a12n-server needs to support OpenID Connect. Time to start testing and see what clients actually need!

The main thing this adds is that we're now emitting the `id_token` which @pschwyter might be familiar with.

### DIFF
--- a/src/oauth2/controller/authorize.ts
+++ b/src/oauth2/controller/authorize.ts
@@ -1,6 +1,6 @@
 import Controller from '@curveball/controller';
 import { Context } from '@curveball/core';
-import { NotFound } from '@curveball/http-errors';
+import { NotFound, NotImplemented } from '@curveball/http-errors';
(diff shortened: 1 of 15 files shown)


src/oauth2/controller/authorize.ts
```diff
@@ -108,16 +117,29 @@ class AuthorizeController extends Controller {
...
+      nonce: params.nonce,
     });
 
+    const redirectParams: Record<string, string> = {
+      code: code.code,
+    };
+    if (params.state) redirectParams.state = params.state;
+
+    if (params.responseType === 'code id_token') {
+      redirectParams.id_token = await generateJWTIDToken({
```
pschwyter, line 129: Not an expert, but I thought the ID token is meant to be included in the token response. Just making sure this isn't an oversight

PR #448, opened 2022-11-10
Fix create group form

evert:


ikbensiep (approved):
I very much believe this is Good:tm:

PR #456, opened 2023-03-29
Fix HTTP/500 error on /authorize endpoint with Sqlite.

evert:
Could have sworn this worked before!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+0.23.1 (????-??-??)
+-------------------
+
+* Fix 500 error on /authorize endpoint when using Sqlite
+
+
 0.23.0 (2022-01-12)
 -------------------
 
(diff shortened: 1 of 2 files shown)


changelog.md
evert, line 1: Ignore this file. I'm using a changelog tool that reformats the changelog a bit.

src/oauth2/service.ts
```diff
@@ -481,7 +481,7 @@ export async function generateAuthorizationCode(options: GenerateAuthorizationCo
       scope: options.scope?.join(' '),
       browser_session_id: options.browserSessionId,
       created_at: Math.floor(Date.now() / 1000),
-      nonce: options.nonce,
+      nonce: options.nonce ?? null,
```
evert, line 484: @ansariswag this fixed this bug. knex now errors when passing undefined. Not sure if this is new behavior :S

ansariswag: @evert , thanks, let me check out and then try.

ansariswag (approved):

Issue #494, opened 2024-03-25
[Docker image] Postgres - no schema has been selected to create in


elaugier:
Hello,

I'm trying to setup a12n-server by using the official docker image in docker compose but i get the following error when the container starts:

```
2024-03-25 11:18:37 Could not start a12n-server
2024-03-25 11:18:37 error: create table "knex_migrations" ("id" serial primary key, "name" varchar(255), "batch" integer, "migration_time" timestamptz) - no schema has been selected to create in
2024-03-25 11:18:37     at Parser.parseErrorMessage (/opt/app/node_modules/pg-protocol/dist/parser.js:287:98)
2024-03-25 11:18:37     at Parser.handlePacket (/opt/app/node_modules/pg-protocol/dist/parser.js:126:29)
2024-03-25 11:18:37     at Parser.parse (/opt/app/node_modules/pg-protocol/dist/parser.js:39:38)
2024-03-25 11:18:37     at Socket.<anonymous> (/opt/app/node_modules/pg-protocol/dist/index.js:11:42)
2024-03-25 11:18:37     at Socket.emit (node:events:518:28)
2024-03-25 11:18:37     at addChunk (node:internal/streams/readable:559:12)
2024-03-25 11:18:37     at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
2024-03-25 11:18:37     at Readable.push (node:internal/streams/readable:390:5)
2024-03-25 11:18:37     at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
2024-03-25 11:18:37   length: 120,
2024-03-25 11:18:37   severity: 'ERROR',
2024-03-25 11:18:37   code: '3F000',
2024-03-25 11:18:37   detail: undefined,
2024-03-25 11:18:37   hint: undefined,
2024-03-25 11:18:37   position: '14',
2024-03-25 11:18:37   internalPosition: undefined,
2024-03-25 11:18:37   internalQuery: undefined,
2024-03-25 11:18:37   where: undefined,
2024-03-25 11:18:37   schema: undefined,
2024-03-25 11:18:37   table: undefined,
2024-03-25 11:18:37   column: undefined,
2024-03-25 11:18:37   dataType: undefined,
2024-03-25 11:18:37   constraint: undefined,
2024-03-25 11:18:37   file: 'namespace.c',
2024-03-25 11:18:37   line: '500',
2024-03-25 11:18:37 ⚾ @curveball/a12n-server 0.25.4
```

[...]

PR #495, opened 2024-03-25
adding public schema in the SEARCH_PATH configuration for pg config

elaugier:
related to the issue [#494 ](https://github.com/curveball/a12n-server/issues/494)

I'm not sure this is best way to solve the issue but by adding the 'public' string in the search path, the knex migration works


evert (approved):

evert:
Thank you!

Issue #519, opened 2024-08-14
Block user accounts if an incorrect password was entered 5 times


evert:
We should have an 'incorrect password' counter per user. If this exceeds a (hardcoded, for now) number we should prevent future attempts to log in.

A successful login should also reset this counter.

This implies that we need a new property in the database on a user, indicating that too many incorrect login attempts were made. This property can just be the number of previous attempts.

Make sure that when this condition happens, the user sees an error like:

    "Too many failed login attempts, please contact [admin] to unlock your account"

PR #527, opened 2024-09-10
Block user accounts if an incorrect password was entered 5 times

YunhwanJeong:
### Feature Overview:
- This PR is an implementation of #519.
- This PR introduces a login activity tracking system that monitors incorrect password attempts per user. If the number of failed attempts exceeds a predefined limit, the user’s account is locked, preventing further login attempts. A successful login resets the failed attempts counter, enhancing account security and preventing unauthorized access.

Screenshot 2024-09-09 at 11 26 04 PM

### Key Changes:
1. Login Activity Tracking:
    - Introduced a new user_login_activity table to store login-related activities, including failed attempts and account lock status.
    - Added a new service (src/login-activity/service.ts) to handle the logic related to tracking failed login attempts and locking accounts.
2. Account Locking Logic:
    - Implemented a maximum failed attempts limit. If the login attempt is made when the account is locked, increment the failed_login_attempts and create log entry with the loginFailedAccountLocked event.
    - If the limit is reached, the account is immediately locked, and a log entry is created with the accountLocked event.
    - Users with locked accounts are presented with an appropriate error message, instructing them to contact the administrator to unlock their account.
3. Resetting Failed Attempts:
    - After a successful login, the failed attempts counter is reset to ensure that future incorrect password attempts are tracked accurately.
4. Controller Enhancements:
    - Updated the post method in the login controller to incorporate the new login activity tracking and account locking logic.
    - Ensured that the resetFailedLoginAttempts function is executed after a successful login to maintain data integrity.
5. Logging:
    - Enhanced logging to include new events such as loginFailedAccountLocked and accountLocked, providing better visibility into account security incidents.

### Security Improvements:
This feature improves the security of user accounts by preventing brute-force attacks and unauthorized access through repeated incorrect password attempts.

### Testing & Validation:
- Manual testing was conducted to validate the correct behavior under various scenarios, including successful logins, incorrect password attempts, and account lockouts.

### Next Steps:
- Should monitor the feature in production to ensure it behaves as expected and doesn’t introduce any performance overhead.
- Consider implementing an admin feature like resetting the failed_login_attempts and account_locked easily.
- Consider implementing additional security features such as IP-based throttling and suspicious login detection in future iterations.

### Checklist
- [x] Performed a self-review.
- [x] Tested manually.
- [ ] Added tests.

### DIFF
--- a/src/log/types.ts
+++ b/src/log/types.ts
@@ -12,6 +12,9 @@ export enum EventType {
 
   oauth2BadRedirect = 11,
   generateAccessToken = 12,
+
+  accountLocked = 13,
+  loginFailedAccountLocked = 14,
 }
 
 export type LogEntry = {
@@ -34,4 +37,6 @@ export const eventTypeString = new Map<EventType, string>([
   [EventType.loginFailedNotVerified,'login-failed-notverified'],
   [EventType.tokenRevoked,          'token-revoked'],
   [EventType.oauth2BadRedirect,     'oauth2-badredirect'],
+  [EventType.accountLocked, 'account-locked'],
+  [EventType.loginFailedAccountLocked, 'login-failed-account-locked'],
 ]);
--- a/src/log/types.ts
+++ b/src/log/types.ts
@@ -13,8 +13,8 @@ export enum EventType {
   oauth2BadRedirect = 11,
   generateAccessToken = 12,
 
-  accountLocked = 13,
-  loginFailedAccountLocked = 14,
+  accountLocked = 13, // Logged at the time the account is locked
+  loginFailedAccountLocked = 14, // Logged when a login attempt is made on a locked account
 }
 
 export type LogEntry = {
--- a/src/login/controller/login.ts
+++ b/src/login/controller/login.ts
@@ -7,7 +7,7 @@ import { EventType } from '../../log/types.js';
 import { MFALoginSession } from '../../mfa/types.js';
 import * as webAuthnService from '../../mfa/webauthn/service.js';
 import { hasUsers, PrincipalService } from '../../principal/service.js';
-import { getSetting, requireSetting } from '../../server-settings.js';
+import { getSetting } from '../../server-settings.js';
 import * as services from '../../services.js';
 import { PrincipalIdentity, User } from '../../types.js';
 import { loginForm } from '../formats/html.js';
@@ -59,24 +59,6 @@ class LoginController extends Controller {
 
     const user = (await principalService.findByIdentity(identity)) as User;
 
-    if (await services.loginActivity.isAccountLocked(user)) {
-      await services.loginActivity.incrementFailedLoginAttempts(user);
-      log(EventType.loginFailedAccountLocked, ctx.ip(), user.id, ctx.request.headers.get('User-Agent'));
-      return this.redirectToLogin(ctx, '', `Too many failed login attempts, please contact ${requireSetting('smtp.emailFrom')} to unlock your account.`);
-    }
-
-    if (!await services.user.validatePassword(user, ctx.request.body.password)) {
-      const incrementedAttempts = await services.loginActivity.incrementFailedLoginAttempts(user);
-
-      if (services.loginActivity.reachedMaxAttempts(incrementedAttempts)) {
-        log(EventType.accountLocked, ctx.ip(), user.id, ctx.request.headers.get('User-Agent'));
-        return this.redirectToLogin(ctx, '', `Too many failed login attempts, please contact ${requireSetting('smtp.emailFrom')} to unlock your account.`);
-      }
-
-      log(EventType.loginFailed, ctx.ip(), user.id);
-      return this.redirectToLogin(ctx, '', 'Incorrect username or password');
-    }
-
     if (!user.active) {
       log(EventType.loginFailedInactive, ctx.ip(), user.id, ctx.request.headers.get('User-Agent'));
       return this.redirectToLogin(ctx, '', 'This account is inactive. Please contact Admin');
@@ -90,7 +72,10 @@ class LoginController extends Controller {
       return;
     }
 
-    await services.loginActivity.resetFailedLoginAttempts(user);
+    const { success, errorMessage } = await services.user.validateUserCredentials(user, ctx.request.body.password, ctx);
+    if (!success && errorMessage) {
+      return this.redirectToLogin(ctx, '', errorMessage);
+    }
 
     ctx.session = {
       user: user,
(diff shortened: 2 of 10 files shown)


src/log/types.ts
```diff
@@ -12,6 +12,9 @@ export enum EventType {
 
   oauth2BadRedirect = 11,
   generateAccessToken = 12,
+
+  accountLocked = 13,
+  loginFailedAccountLocked = 14,
```
evert, line 17: I think it's worth adding some comments to these 2 lines so it's a bit clearer what the difference is.

src/login/controller/login.ts
```diff
@@ -56,9 +56,23 @@ class LoginController extends Controller {
...
+
+    const user = (await principalService.findByIdentity(identity)) as User;
+
+    if (await services.loginActivity.isAccountLocked(user)) {
+      await services.loginActivity.incrementFailedLoginAttempts(user);
+      log(EventType.loginFailedAccountLocked, ctx.ip(), user.id, ctx.request.headers.get('User-Agent'));
+      return this.redirectToLogin(ctx, '', `Too many failed login attempts, please contact ${requireSetting('smtp.emailFrom')} to unlock your account.`);
+    }
 
     if (!await services.user.validatePassword(user, ctx.request.body.password)) {
```
evert, line 68: This is a bit of a larger request, but I wonder if _most_ of this logic should go into the service. The reason is that this is not the only place where we check a username and password, there are two more places:

* OAuth2 has a `password` flow
* I recently added a 'authorization_challenge` OAuth2 flow.

These 3 features all do a username/password check so it kinda makes sense that they all increment the 'failed login' counter if a password fails.

I'm not sure if this should just be part of 'validatePassword' or a new function.

src/login/login-activity/service.ts
```diff
@@ -0,0 +1,65 @@
+import { UserLoginActivityRecord } from 'knex/types/tables.js';
+import db from '../../database.js';
+import { User } from '../../types.js';
+
+const MAX_FAILED_ATTEMPTS = 5;
+
+export function reachedMaxAttempts(attempts: number) {
+  return attempts === MAX_FAILED_ATTEMPTS;
```
evert, line 8: I think this should be `>=` instead of `===`

src/login/login-activity/service.ts
```diff
@@ -0,0 +1,65 @@
+import { UserLoginActivityRecord } from 'knex/types/tables.js';
```
evert, line 1: Is `login-attempts` a better name for this?

src/login/login-activity/service.ts
```diff
@@ -0,0 +1,65 @@
...
+import db from '../../database.js';
+import { User } from '../../types.js';
+
+const MAX_FAILED_ATTEMPTS = 5;
+
+export function reachedMaxAttempts(attempts: number) {
+  return attempts === MAX_FAILED_ATTEMPTS;
+}
+
+async function getLoginActivity(user: User): Promise<UserLoginActivityRecord | undefined> {
```
evert, line 11: Instead of returning `undefined`, how about always returning a record? Because maybe if there's no database record yet, you can guess what the record should be for the initial state

src/login/login-activity/service.ts
```diff
@@ -0,0 +1,65 @@
...
+      account_locked: 0,
+    })
+    .onConflict('principal_id')
+    .ignore();
+}
+
+export async function incrementFailedLoginAttempts(user: User): Promise<number> {
+  await ensureUserLoginActivityRecord(user);
+
+  return await db.transaction(async (trx) => {
```
evert, line 31: :+1:

evert (commented):
Thank you, this is a great contribution!

Happy to chat about my feedback. They're loosely held opionions

src/login/controller/login.ts
YunhwanJeong: Updated: 33288d2629c953dbb59581a631739c57e04cf7d8
Could you double-check the use cases of OAuth2 password flow and the OAuth2 authorization_challenge flow?

YunhwanJeong (commented):
I've incorporated your review! @evert

src/login/service.ts
```diff
@@ -228,21 +230,11 @@ async function challengeUsernamePassword(session: LoginSession, username: string
...
-      session,
-      'Incorrect username or password',
-      'username-password',
-      true,
-    );
-  }
-
   session.principalId = user.id;
   session.passwordValid = true;
   session.dirty = true;
```
evert, line 235: For context, this is a new service I've been working on, and the goal is to eventually handle all login flows in a consistent way including all the special cases like TOTP, WebAuthN and other MFA flows.

The issue here is that this service is very imperative, so it takes one step at the time to 'build up' the entire login session and ask for all the credentials/factors.

Moving the login check later in this function means that you set 'passwordValid' to true before the actual check is done. This information gets saved in the 'login session', so even if the password was wrong, you end up with a session that no longer requires a password check. See line 240-242 in the old diff and 233-235 in the new one.

This is actually the flow we use for sprout now. Once this is fixed I think this is good to go!

YunhwanJeong: @evert Updated: 3a52f2fc2409d9c0bd0cd38bfa216db619dbb259

For context, I moved this logic because the `validateUserCredentials` method calls `await loginActivityService.resetFailedLoginAttempts(user);`. My intention was to delay the execution of `validateUserCredentials` method as much as possible so that `resetFailedLoginAttempts` would run only when the login actually succeeds.

However, in this case, that approach is not feasible because as you mentioned, we need to execute the `validateUserCredentials` method before setting `session.passwordValid = true;`. Thank you for pointing that out!

I also considered running the `resetFailedLoginAttempts` method imperatively in each controller/service. However, I believe the current approach is fine since we call the `validateUserCredentials` method just before logging the `loginSuccess` event in all controllers/services, except for this one. And it makes sense to run the `resetFailedLoginAttempts` method after checking all relevant aspects in the `validateUserCredentials` method,  such as `isAccountLocked`, `validatePassword` or `reachedMaxAttempts`, even while other checks for login are still ongoing.

evert:
Thank you so much! Hope that wasn't too painful

evert (approved):

src/login/service.ts
evert: This works for me! Thank you. Login Service is still a work in progress so I probably need to spend some more time making it's logic a bit more clear

src/login/controller/login.ts
evert: Gonna do this while working on #513

PR #533, opened 2024-10-02
Fix(postgres): Fix postgres server start and type cast `created_at` field

Zen-cronic:
Fixed some compatibility issues with postgres. These issues don't pop up with `mysql`. 

Pls see review for detailed comments.


src/database.ts
```diff
@@ -132,6 +132,7 @@ export function getSettings(): Knex.Config {
         searchPath = [
           connection.user as string,
           connection.database as string,
+          'public',
```
Zen-cronic, line 135: Added missing default schema.

src/principal/service.ts
```diff
@@ -411,8 +411,8 @@ function recordToModel(user: PrincipalsRecord): Principal {
     href: `/${userTypeIntToUserType(user.type)}/${user.external_id}`,
     externalId: user.external_id,
     nickname: user.nickname!,
-    createdAt: new Date(user.created_at),
-    modifiedAt: new Date(user.modified_at),
+    createdAt: new Date(+user.created_at),
+    modifiedAt: new Date(+user.modified_at),
```
Zen-cronic, line 415: Both fields are stored as BigInt in postgres. But when `knex` (or upstream `pg`) queries that data, a string version of the number (e.g., `"179289..."`) is returned. 

This may not be an issue for mysql and sqlite. If that's the case the `+` is a no-op.

evert: Thanks for finding this. This is going to be annoying going forward with postgres =(

Issue #534, opened 2024-10-03
When changing password, a 403 CSRF error is thrown


Zen-cronic:
Flow:

1. Log in as a user or admin
2. Visit the change-password [page](https://github.com/curveball/a12n-server/blob/e855ee60ae7f649622937a3abaab8f7e592b6116/templates/changepassword.hbs).
3. Correctly complete the input form. 
4. `403` [CsrfError](https://github.com/curveball/session/blob/e8bc60ec1b6560dd14fc334fa392dc991976b648/src/mw.ts#L60) returned. 

Related file:

- csrf [middleware](https://github.com/curveball/a12n-server/blob/e855ee60ae7f649622937a3abaab8f7e592b6116/src/middleware/csrf.ts)
Here, `ctx.session['csrf-token']` exists but it's not passed to `ctx.validateCsrf()`.

PR #535, opened 2024-10-03
Fix(csrf): Send csrf token for the `/change-password` path

Zen-cronic:
For #534.

A `csrf-token` already exists on `ctx.session`. But we need to pass it to `validateCsrf()`.

### DIFF
--- a/src/middleware/csrf.ts
+++ b/src/middleware/csrf.ts
@@ -10,7 +10,7 @@ const safePaths = [
 
 export default function(): Middleware {
 
-  return (ctx, next) => {
+  return async (ctx, next) => {
 
     /**
      * There's 2 ways a user might be authenticated, via a session cookie or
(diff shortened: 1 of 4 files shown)


src/middleware/csrf.ts
```diff
@@ -21,7 +21,12 @@ export default function(): Middleware {
     if (!ctx.session.user) return next();
 
     if (!safeMethods.includes(ctx.method) && !safePaths.includes(ctx.path)) {
-      ctx.validateCsrf();
+      if(ctx.path === '/change-password'){
+        ctx.validateCsrf(await ctx.getCsrf());
```
evert, line 25: This change unfortunately doesn't really make sense. `getCsrf()` returns the current CSRF token, so this will always be valid.

The point of `validateCsrf` is to validate a token that was submitted with the request. The solution here would be to ensure that the HTML form has an `<input type="hidden" name="csrf-token" value"..." />` included.

Zen-cronic: Gotcha. B/c the current form doesn't have the csrf-token value, I'll try this:

1. Send csrf-token from the `/change-password` mw.
2. Attach csrf-token in `changepassword.hbs`.
3. Validate the token back in the server.

Zen-cronic: Fixed

evert: Excellent!

evert:
Thank you so much for this!

Zen-cronic:
no problem!

PR #556, opened 2024-12-10
fix(bin): Update cli script to ESM

### LINKED ISSUE #555: Cli script should follow ESM
Running `npx @curveball/a12n-server` throws `ReferenceError: __dirname is not defined in ES module scope` due to [this](https://github.com/curveball/a12n-server/blob/f8ef1c7898959e8cd30d110805e2ebf898b021c7/bin/a12n-server#L12).

Fix: get the dirname this way https://github.com/curveball/a12n-server/blob/f8ef1c7898959e8cd30d110805e2ebf898b021c7/src/env.ts#L15

Zen-cronic:
Fixes #555

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "README.md",
     "assets",
     "dist",
+    ".env.defaults",
     "package.json",
     "schemas",
     "src",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "README.md",
     "assets",
     "dist",
+    ".env.defaults",
```
Zen-cronic, line 29: Without this, `npx @curveball/a12n-server` doesn't have a source file to copy from (given that there isn't already a .env file)

evert: Thanks! wild how broken this was =(

PR #579, opened 2025-01-28
Enhance/add email verification challenge

chelsearoman-ca:


src/login/challenge/email-verification.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+  * Checks if the user has an unverified email.
+  * Returns true if an unverified email identity is found,
+  */
+
+  async userHasChallenge(): Promise<boolean> {
+
+    const identity = await this.findUnverifiedEmailIdentity();
+    return identity !== null;
```
evert, line 32: So one way this Challenge has to behave slightly different from the other one, is that this should check the _specific_ identity (username) that the user used to login, whereas the email-otp will accept any.

The problem you have here is that you don't have access to this in this function. I think the AbstractLoginChallenge class should be modified to add 1 more constructor argument, so the signature becomes:

```
constructor(principal: User, identity: PrincipalIdentity, logger: UserEventLogger, ip: string) {
```

Then `userHasChallenge` can operate on `this.identity`.

This will then make a few helper functions in this class also unneeded.

The logic is that any account can have 1 or more identities associated to it, but only a verified identity can be used to log in.

src/login/challenge/email-verification.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      // Mark the identity as verified
+      await markVerified(identity);
+      return true;
+    } catch (err) {
+      if (err instanceof BadRequest) {
+        await services.principalIdentity.sendVerificationRequest(identity, this.ip);
+        throw new A12nLoginChallengeError(
+          'Invalid or expired email_verification',
+          'email_not_verified',
```
evert, line 60: Lets introduce a new error code: `email_verification_code_invalid`

evert (changes requested):

evert (approved):
Ship it!

PR #586, opened 2025-02-01
OICD configuration endpoint

evert:
A bunch of this was already written, just needed a route!

### DIFF
--- a/src/login/formats/html.ts
+++ b/src/login/formats/html.ts
@@ -33,3 +33,17 @@ export function mfaForm(
   }, 'minimal-form');
 
 }
+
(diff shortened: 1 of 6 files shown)


usrrname (approved):

src/login/formats/html.ts
```diff
@@ -33,3 +33,17 @@ export function mfaForm(
   }, 'minimal-form');
 
 }
+
+export function experimentalLoginForm(msg: string, error: string, hiddenFields: KeyValue, registrationEnabled: boolean, registrationUri: string, resetPasswordUri: string): string {
```
usrrname, line 37: Semantic nitpick: what is experimental about this login?

evert: did not mean to commit this! was from a WIP branch. I'm working on a different way to handle the login flow that handles MFA a bit better. I figure I just keep this stuff under a different route until it's done and can replace the regular login. For now I'll just uncommit this.

PR #587, opened 2025-02-01
doc updates

usrrname:
# Changes


docs/authorization-code-flow.md
```diff
@@ -0,0 +1,89 @@
+# Authorization Code Flow
+
+The OAuth 2.0 Authorization Code Flow is one of three grant types of OAuth2 that enables an application to access data on a user's behalf.
+This allows an application to selectively receive more sensitive, opt-in information about a user.
+
+For more detailed information about the Authorization Code Flow, see the [OAuth 2.0 Authorization Framework RFC]().
+
+Currently, `a12n-server`'s' authorization code flow provides access tokens for the following requests:
+
```
usrrname, line 9: @evert wanna add examples of requests that can be made at endpoints?

ie. POST/PUT/GET /<endpoint>

evert: I think I probably want to steer people away a bit from having to learn about OAuth2 works. Resources are out there, and what we really want is people using standard clients so they _don't_ have to learn.

I imagine you probably got to the same conclusion after going through the next-auth gauntlet, but let me know!

usrrname: I feel several ways about this (discussion below.)

I think we (all contributors) need to be clear on:

1. Audience: Who are audiences for this tool, and what is their skill level? 

2. Motivation: What gap a12n-server tries to fill in the existing auth tool landscape. What problem does it try to solve? 

3. What is a12n-server responsible for vs regular server development or full stack dev of an auth flow? 
(I know some of this we need to find out as I attempt integrations) 

My impression: the regular dev-as-user for this wouldn't be a newcomer to dev... or would already know about comparable clients and services even if they're unfamiliar with auth. 

I also imagined such a regular dev might want  enough context such that they can modify/extend the server behaviour or know what to expect without debugging too hard if they want to customize something. So far I gather the only place they do config is the `.env`?

2. I copied some of the format of this doc from Battlenet (amongst other existing solutions which seem to offer short explainers and how their APIS work)

https://develop.battle.net/documentation/guides/using-oauth

https://develop.battle.net/documentation/guides/using-oauth/authorization-code-flow 

Let me know what you think... I'm ok with removing this now and gathering more info about common use-cases and adding them as needed.


Some of this was hopeful layup doc work for the university crowd who is gonna want to CRUD to a a12n-server, or perhaps you have something else in mind?

PR #589, opened 2025-02-03
Discovery document should use all absolute URLS.

evert:
A client that shall not be named requires it.

### DIFF
--- a/schemas/app-client-edit-form.json
+++ b/schemas/app-client-edit-form.json
@@ -9,34 +9,33 @@
   "additionalProperties": false,
 
   "properties": {
+    "clientId": {
+      "type": "string",
+      "description": "client_id for this client. May be omitted for edits and must not change from the original",
+      "readOnly": true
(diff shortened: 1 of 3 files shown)


schemas/app-client-edit-form.json
```diff
@@ -9,34 +9,33 @@
   "additionalProperties": false,
```
evert, line 9: unrelated bugfix :P

src/well-known/formats/json.ts
```diff
@@ -43,15 +42,17 @@ type MetaData = {
 
 export function metadata(): MetaData {
 
+  const origin = getGlobalOrigin();
+
   return {
-    issuer: getGlobalOrigin(),
-    authorization_endpoint: '/authorize',
+    issuer: origin,
+    authorization_endpoint: origin + '/authorize',
```
evert, line 50: this is where it went wrong

PR #600, opened 2025-02-03
fix(package.json): adds `@curveball/static` as dependency on `package.json`

usrrname:
# Changes

When I run `pnpm install` and then `pnpm run build` I get the following error:
```
> make build

./node_modules/.bin/tsc
src/routes.ts:2:22 - error TS2307: Cannot find module '@curveball/static' or its corresponding type declarations.

2 import staticMw from '@curveball/static';
                       ~~~~~~~~~~~~~~~~~~~


Found 1 error in src/routes.ts:2

make: *** [dist/build] Error 2
 ELIFECYCLE  Command failed with exit code 2.
```

If `@curveball/a12n-server` is installed with `npm` `@curveball/static` is in the `package-lock.json`, but this won't resolve in `pnpm-lock.yaml` in a workspace.

This PR makes the dependency available on `package.json` so those using pnpm will be able to build the package.

so it will resolve on `pnpm install`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@curveball/problem": "^1.0.0",
         "@curveball/router": "^2.0.0",
         "@curveball/session": "^1.0.0",
+        "@curveball/static": "^1.0.0",
         "@curveball/validator": "^1.0.0",
         "@simplewebauthn/server": "^10.0.0",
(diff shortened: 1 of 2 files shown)


evert:
ty! was def an oversight

PR #601, opened 2025-02-03
fix(env.defaults): remove trailing slash from `CURVEBALL_ORIGIN`

usrrname:
# Changes

fix: removes trailing slash from `.env.defaults` `CURVEBALL_ORIGIN` so that they don't end up in JSON uris

Pasted_Image_2025-02-03__8_12 AM

### DIFF
--- a/.env.defaults
+++ b/.env.defaults
@@ -12,7 +12,7 @@ KEEP_ALIVE_TIMEOUT_MS=
 # Public URL. This must be the URL that users will use to access a12nserver.
 # If this is not provided or incorrect, this will result in the server not
 # functioning correctly.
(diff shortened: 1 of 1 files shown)


.env.defaults
```diff
@@ -12,7 +12,7 @@ KEEP_ALIVE_TIMEOUT_MS=
 # Public URL. This must be the URL that users will use to access a12nserver.
 # If this is not provided or incorrect, this will result in the server not
 # functioning correctly.
-CURVEBALL_ORIGIN="http://localhost:8531/"
+CURVEBALL_ORIGIN="http://localhost:8531"
```
evert, line 15: Great tnx!

evert (approved):

PR #604, opened 2025-02-04
setup(package.json): adds `prepare` script for easy `.env` generation

usrrname:
# Changes

- setup(package.json): adds prepare script for easy `.env` generation
- adds doc telling people to use it in "Getting Started"

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,7 +1,11 @@
-{:toc}
-
 Getting started
 ===============
+
+Contents:
+
+* [The dev server](#the-dev-server)
+* [Deploying in production](#deploying-in-production)
+* [Running from source](#running-from-source)
 
 
 The dev server
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -8,102 +8,101 @@
   - [Running the server](#running-the-server)
   - [Creating the first user](#creating-the-first-user)
 
+
 Getting started
 ===============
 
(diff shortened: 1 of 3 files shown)


usrrname:
I'm now wondering if the package.json commands are run in order of appearance as tests are breaking ...

docs/getting-started.md
```diff
@@ -33,6 +33,14 @@ cd a12n-server
...
 ```
 
+Setup the environment:
+
+```sh
+npm run prepare
+```
+
+This will create a `.env` file with a private JWT for local development of OAuth2/OIDC flows.
+
```
evert, line 43: This probably needs a bit of work first. `prepare` is a bit of a misnomer, because the script actually is the 'dev server' script, also mentioned in `README.md`, so it does more than just preparing, it actually starts the server with default (dev) configuration. But I agree that we should make it easier to get started.

I'm thinking we really 4 separate documents for each use-case
 
1. dev server for testing. (`npx script`)
2. from source (requires a checkout, `npm i`, etc).
3. in production

And possibly also :cry: 

4. With docker compose

usrrname: yeah so far, getting-started is for developers who are using this package, not for contributors right?

usrrname: When you say "from source", you really mean it's the doc for contributors right?

evert: yes!

PR #618, opened 2025-02-09
Test if docker will build for every PR

evert:


Dockerfile
```diff
@@ -1,21 +1,20 @@
 # Stage 1: build
-FROM node:20-alpine as build-stage
+FROM node:22-alpine as build-stage
 WORKDIR /opt/app
 
-# Needed for building @vscode/sqlite3 package
-RUN apk add python3 make gcc musl-dev g++
+# Needed for building binary Node.js addons
+RUN apk add python3 make gcc musl-dev g++ py3-pip py3-setuptools
```
evert, line 6: This was the only thing that was needed in the end.

Issue #624, opened 2025-02-12
Add fields to user resources


evert:
Users resources should have the following fields attached to it (all optional)

* givenName
* middleName
* familyName
* birthDate
* address (JSON type with fields `address: string[], locality, region, postalCode, country`)
* locale (such as `en-CA`)
* zoneinfo (such as `Europe/Amsterdam`)
* metadata (type Record<string, string>`)

Most of these are described here: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

usrrname:
Sooo I got confused about this one in terms of which areas to update in the next task to support fields in GET /user:id

It seems like I need to do something about:

1. Insert `user_info` records into the db when user login happens 
1. Create an `OidcUserinfoService` or piggyback on the PrincipalService to find userinfo by id 

Does that sound right?

evert:
> Insert user_info records into the db when user login happens

No this is not needed. Inserting / updating should (for now) only happen with a `PUT` request on a user resource.

> Create an OidcUserinfoService or piggyback on the PrincipalService to find userinfo by id

Yes, it should be either a new `userinfo/service.ts` or on `user/service.ts`, but this is _not_ a OpenID-connect-specific feature. We're just using OpenID terminology. (we do want to add this to the OpenID Connect userinfo endpoint but only after the core feature is done).

I'll edit the task list

PR #631, opened 2025-02-19
#624 feat: `user_info` table

usrrname:
# Changes

#624 
- creates `user_info` table schema

Source: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims


src/migrations/043_userinfo_oidc.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  if (result.length) {
+    // Migration has been applied using the old patch system
+    return;
+  }
+
+  await knex('changelog').insert({
+    id: 43,
+    timestamp: Math.floor(Date.now() / 1000)
+  });
```
evert, line 15: All this stuff is for the oldest migrations, before Knex was used. These (double-digit) old migrations live in here so there's an ugrade path.

This migration should use the timestamp format so it happens at the very end.

src/migrations/043_userinfo_oidc.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  await knex.schema.createTable('user_info', table => {
+    table.increments();
+    table.string('sub', 50).notNullable();
+    table.string('name', 200).nullable();
+    table.string('email', 100).nullable();
+    table.boolean('email_verified').nullable();
```
evert, line 22: I kind of went into this in #624, but only a few fields are actually needed. A lot of these fields already in the database in some form, so you can remove everything that wasn't already listed in #624, and it would be nice to add `metadata` for custom fields

src/migrations/043_userinfo_oidc.ts
```diff
@@ -0,0 +1,56 @@
...
+    table.string('website', 300).nullable().checkRegex('^(http://|https://).*$');
+    // TODO: Check for right way to add JSON properties
+    table.json('address').defaultTo({
+      'formatted': '',
+      'street_address': '',
+      'locality': '',
+      'region': '',
+      'postal_code': '',
+      'country': ''
+    });
```
evert, line 44: I don't think `json` works consistently across the different databases we support, so I would suggest just using `text` here.

`default` doesn't work in sqlite, so it's best to remove it.

src/migrations/043_userinfo_oidc.ts
usrrname: > A lot of these fields already in the database in some form

Should I reference them with... refs, queries or foreign keys?

evert: No I meant check out the ticket and remove all fields that aren't mentioned there. We don't want data to be in our database duplicated

usrrname: Ohh ok! a schema is not like a query view! lol

evert: No it's a database table

src/migrations/043_userinfo_oidc.ts
```diff
@@ -0,0 +1,34 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+
+  const result = await knex('changelog').select('id').where({ id: 43 });
+
+  if (result.length) {
+    // Migration has been applied using the old patch system
+    return;
+  }
```
evert, line 10:
```suggestion
```

src/migrations/043_userinfo_oidc.ts
```diff
@@ -0,0 +1,34 @@
+import { Knex } from 'knex';
```
evert, line 1: This migration filename still needs a more 'recent' number. Check out the patch directory, most of the later ones all use a timestamp instead of a 3 digit number.

src/migrations/043_userinfo_oidc.ts
```diff
@@ -0,0 +1,34 @@
...
+        table.string('name', 200).nullable()
+        table.string('middle_name').nullable()
+        table.string('given_name', 30).nullable()
+        table.string('family_name', 30).nullable()
+        table.date('birthdate').nullable()
+        table.string('picture').nullable()
+        table.integer('updated_at').unsigned().nullable()
+        table.text('address').nullable()
+        table.string('locale', 5).nullable()
+        table.string('zoneinfo').nullable()
```
evert, line 24: Just a thought, but in many database types we can add comments to fields, like this patch:

https://github.com/curveball/a12n-server/blob/main/src/migrations/20240908210700_block_failed_login_attempts.ts

The benefit of having comments on fields is when we auto-generate the db-types.ts file, these comments also get added to the types like here:

https://github.com/curveball/a12n-server/blob/main/src/db-types.ts#L221

It's not strictly needed so entirely up to you. As you can see in db-types.ts, I only started doing this for later migrations.

src/migrations/043_userinfo_oidc.ts
usrrname: oh hey I missed this! yeah I made a commit while waiting for something at work

src/migrations/043_userinfo_oidc.ts
usrrname: oh yeah I liked those! 

I was also wondering if we should keep picture as I saw a ton of examples wit hthat

evert: I *love* the idea of having a picture, but maybe it's just a separate ticket. It's not really as simple as we need to have a way to actually change the picture via an upload. Uploads can be malicious, and we need to resize probably so there's just a bit more to think about.

I kept out `gender` btw because i would rather use `pronouns` which has the added benefit of also not being usable by the US government.

usrrname: yeah I saw the spec... only male and female 😬

usrrname: I ex_spec_t spec to be stuffy and outdated tho

evert (approved):
lgtm to me if `lint` passes and `picture` is removed.

src/migrations/20251740066033_user_info.ts
```diff
@@ -0,0 +1,26 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+
+  await knex.schema.createTable('user_info', table => {
+    table.increments();
+    table.string('sub', 50).notNullable().comment('Subject - Identifier for the End-User at the Issuer.');
+    table.string('name', 200).nullable().comment('End-User\'s full name in displayable form including all name parts, possibly including titles and suffixes, ordered according to the End-User\'s locale and preferences.');
+    table.string('middle_name').nullable().comment('Casual name of the End-User that may or may not be the same as the given_name.');
```
evert, line 9: This comment mixes up `nick_name`, middle name is just the second name.

src/migrations/20251740066033_user_info.ts
```diff
@@ -0,0 +1,26 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+
+  await knex.schema.createTable('user_info', table => {
+    table.increments();
+    table.string('sub', 50).notNullable().comment('Subject - Identifier for the End-User at the Issuer.');
```
evert, line 7: Delete this one. We already have ids for users.

src/migrations/20251740066033_user_info.ts
```diff
@@ -0,0 +1,26 @@
...
+    table.string('name', 200).nullable().comment('End-User\'s full name in displayable form including all name parts, possibly including titles and suffixes, ordered according to the End-User\'s locale and preferences.');
+    table.string('middle_name').nullable().comment('Casual name of the End-User that may or may not be the same as the given_name.');
+    table.string('given_name', 30).nullable().comment('Given name(s) or first name(s) of the End-User.');
+    table.string('family_name', 30).nullable().comment('Surname(s) or last name(s) of the End-User.');
+    table.date('birthdate').nullable().comment('End-User\'s birthday');
+    table.integer('updated_at').unsigned().nullable().comment('Time the End-User\'s information was last updated.');
```
evert, line 13: This should be `modified_at` for consistency with other tables.

Might also be nice to add `created_at`. Most other tables also have this.

src/migrations/20251740066033_user_info.ts
usrrname: Oooh great catch!

src/migrations/20251740066033_user_info.ts
usrrname: Ooh should it be that, or should I follow the spec? https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims:~:text=Section%C2%A05.1.1.-,updated_at,-number

usrrname: `created_at` added... always wonder how people get that exact time right :p

src/migrations/20251740066033_user_info.ts
```diff
@@ -0,0 +1,26 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+
+  await knex.schema.createTable('user_info', table => {
+    table.increments();
```
evert, line 6: Sorry last one! We don't need an auto_incrementing field here, but we _do_ need:

```suggestion
    table.integer('principal_id').notNullable().primary().references('principals.id')
```

Users are stored in the `principals` table. A principal is either a user, an app or a group.

So we need to re-use the primary key, which is also a foreign key on this table.

src/migrations/20251740066033_user_info.ts
evert: It should be `modified_at`, because this server doesn't just support OIDC. It's own internal data-model should be consistent with itself. So this also uses camelCase in the API for example.

But when we need to encode `modified_at` for OpenID response bodies, it does get changed to `updated_at`. I think if I started from 0 I would probably go for `updated_at` everywhere because it seems a bit more common, but all tables right now use `modified_at`.

evert: So I kind of see a12n-server in the first place be consistent with itself, but then if we need to implement any standard or protocol we do convert to whatever they want.

src/migrations/20251740066033_user_info.ts
usrrname: that makes sense to me for the purposes of table having an id for the user... I also had something similar in my initial PR

src/migrations/20251740066033_user_info.ts
usrrname: Oh yeah, I was wondering... when to use camelCase and when to use underscore 😕

src/migrations/20251740066033_user_info.ts
```diff
@@ -0,0 +1,26 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+
+  await knex.schema.createTable('user_info', table => {
+    table.integer('principal_id').notNullable().primary().references('principals.id');
```
evert, line 6: It looks like the migration failed, but only on MySQL. I suspect this is just missing `unsigned()` for MySQL to pass. Can be merged after!

evert (approved):

PR #636, opened 2025-02-20
Feat: Add users collection pagination

### LINKED ISSUE #625: Paging for the user collection
The user collection is not paged, and this can become slow when there's a ton of users.

Paging should be added with a standard page size of 100. Paging should be implemented using HAL next/previous links.

Zen-cronic:
Fixes #625 

### Changes 
- Update knex query for pagination
- Add tests for `hal.collection`


src/user/formats/hal.ts
```diff
@@ -4,12 +4,22 @@ import { HalResource } from 'hal-types';
...
+
+  const totalUsers = users.length;
+  const totalPages = Math.ceil(totalUsers / pageSize);
+  const nextPage = currentPage < totalPages ? currentPage + 1 : null;
+  const prevPage = currentPage > 1 ? currentPage - 1 : null;
+
+  const startIdx = (currentPage - 1) * pageSize;
+  const endIdx = Math.min(startIdx + pageSize, totalUsers);
+
+  const paginatedUsers = users.slice(startIdx, endIdx);
```
Zen-cronic, line 17: This logic is redundant b/c the db query is already paginated. Only included here for the current tests. 
To Be Removed.

src/user/formats/hal.ts
```diff
@@ -20,9 +30,19 @@ export function collection(users: User[], embeddedUsers: HalResource[]): HalReso
...
+    total: totalUsers,
+    currentPage,
+    totalPages,
   };
 
+  if(nextPage){
+    hal._links['next'] = { href: `/user?page=${nextPage}` };
+  }
+
+  if(prevPage){
```
Zen-cronic, line 42: `next` and `previous` are added only if pagination is necessary.

src/principal/service.ts
```diff
@@ -46,20 +46,35 @@ export class PrincipalService {
...
     }
 
-    const result = await db('principals')
-      .where(filters);
+    let result: PrincipalsRecord[] = [];
+
+    if(type && page !== undefined){
+
+      page = page < 1 ? 1 : page;
+      const pageSize = 100;
```
Zen-cronic, line 66: TODO: store as a constant (it's used elsewhere)

src/user/controller/collection.ts
```diff
@@ -11,7 +11,11 @@ class UserCollectionController extends Controller {
   async get(ctx: Context) {
 
     const principalService = new services.principal.PrincipalService(ctx.privileges);
-    const users = await principalService.findAll('user');
+
+    const pageInt = parseInt(ctx.request.query.page);
+    const page = isNaN(pageInt) ? 1 : pageInt;
```
evert, line 16: Alternative (if it's not too code-golfy)


```suggestion
    const page = +ctx.request.query.page || 1;
```

evert (commented):
great progress!

src/principal/service.ts
```diff
@@ -46,20 +46,35 @@ export class PrincipalService {
...
     const filters: Record<string, any> = {};
     if (type) {
       filters.type = userTypeToInt(type);
     }
 
-    const result = await db('principals')
-      .where(filters);
+    let result: PrincipalsRecord[] = [];
+
+    if(type && page !== undefined){
```
usrrname, line 63: Could also be written as: 

```suggestion
    if (type && page){
```

Because `undefined` is falsy

usrrname (commented):
Ahh rad you added tests!!

src/principal/service.ts
Zen-cronic: oh, but `0` is also falsy, but that's valid input

src/user/controller/collection.ts
evert: No I don't super care!

Zen-cronic:
> Ahh rad you added tests!!

as always :))
ngl, this repo warrants wayy more tests - both unit & e2e

evert:
Awesome, this is looking great!

Zen-cronic:
both previous and next buttons verified:

2nd page of total 3 pages:
three-pages-second

src/user/controller/collection.ts
usrrname: I'm also a `Number()`fan though I get `+`

src/database.ts
```diff
@@ -3,6 +3,8 @@ import { Knex, default as knex } from 'knex';
 import * as path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import './env.ts';
+import dotenv from 'dotenv';
+dotenv.config({ path: './.env.test' });
```
Zen-cronic, line 7: TODO: although the tests pass, this top-level config messes up the dev app (i.e., the dev app should be run via the default `a12nserver.sqlite3` file.)

test/pagination/services.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import * as dotenv from 'dotenv';
+dotenv.config({ path: './.env.test'});
+
+import { PrincipalService, recordToModel } from '../../src/principal/service.ts';
+import db, { insertAndGetId } from '../../src/database.ts';
+
+describe('findAll users pagination', () => {
+
+  const userRecordsDb: unknown[] = [];
```
Zen-cronic, line 12: TODO: use `PrincipalsRecord` type

src/database.ts
Zen-cronic: without this, the tests use the same db for development. we instead need a fresh db for testing as specified in the `.env.test` config file

Makefile
```diff
@@ -17,7 +17,7 @@ docker-run:
 	docker run -it --rm --name $(DOCKER_IMAGE_NAME)-01 $(DOCKER_IMAGE_NAME)
 
 test:
-	npx tsx --test ${TEST_FILES}
+	NODE_ENV=test npx tsx --test ${TEST_FILES}
```
Zen-cronic, line 20: workaround for `src/database.ts` to use the default `.env` file in dev mode and `.env.test` in test mode

Zen-cronic: explanation: https://github.com/curveball/a12n-server/pull/636#discussion_r1964539727

Zen-cronic:
First and last page links implemented:

On the 1st page:

first-page-first-last

On the last page:

last-page-first-last

On a middle page:

middle-page-first-last

Makefile
evert, line 20: I have another suggestion for this:

```suggestion
	DB_DRIVER=sqlite DB_FILENAME=":memory" npx tsx --test ${TEST_FILES}
```

I think this will work as well. Slightly prefer this right now because I want to do something more sophisticated a little later.

src/principal/service.ts
```diff
@@ -46,26 +47,57 @@ export class PrincipalService {
 
   }
 
-  async findAll(type: 'user'): Promise<User[]>;
+  async findAll(type: 'user', page: number): Promise<PaginatedResult<User>>;
   async findAll(type: 'group'): Promise<Group[]>;
   async findAll(type: 'app'): Promise<App[]>;
   async findAll(): Promise<Principal[]>;
-  async findAll(type?: PrincipalType): Promise<Principal[]> {
+  async findAll(type?: PrincipalType, page?: number): Promise<Principal[] | PaginatedResult<Principal>> {
```
evert, line 54: I love the idea of the PaginatedResult, but I think this function should either:

* Always return that type
* Never return it, and instead we create a new function `search()`

I'm partial to keeping `findAll` for simple cases, and adding a new `search` which can be used as a foundation later to do other kinds of filtering.

src/principal/service.ts
```diff
@@ -46,26 +47,57 @@ export class PrincipalService {
...
     }
-    return principals;
+
+    if(type && page !== undefined){
+      return {
+        principals,
+        total,
+        page,
+        pageSize,
+        hasNextPage,
```
evert, line 96: Big fan of this pattern.

src/types.ts
```diff
@@ -155,6 +155,17 @@ export type PrincipalStats = {
...
+/**
+ * Paginated Result
+ */
+export type PaginatedResult<T> = {
+  principals: T[];
+  total: number;
+  page: number;
+  pageSize: number;
+  hasNextPage: boolean;
+}
```
evert, line 167: Given the name of this type, I think this can use a small tweak to be useful for other lookups that aren't principals:

```suggestion
export type PaginatedResult<T> = {
  items: T[];
  total: number;
  page: number;
  pageSize: number;
  hasNextPage: boolean;
}
```

src/user/formats/hal.ts
```diff
@@ -1,14 +1,18 @@
...
+export function collection(embeddedUsers: HalResource[], paginatedResult: PaginatedResult<User>): HalResource {
+
+  const { principals: users, page: currentPage, total, pageSize, hasNextPage } = paginatedResult;
+
+  const totalPages = Math.ceil(total / pageSize);
 
   const hal: HalResource = {
     _links: {
-      'self': { href: '/user' },
+      'self': { href: `/user?page=${currentPage}` },
```
evert, line 15: Could generating the string `/user=${currentPage}` be done in a new function such as:

```javascript
function getUserPageHref(result: PaginatedResult<User>): HalLink {
  if (result===1) return { href: '/user' };
  return { href: `/user?page=${currentPage}` };
}
```

This ensures that we're always consistent with generating the link, and the `self` link lines up with the URL which is useful for caching.

test/pagination/services.ts
```diff
@@ -0,0 +1,85 @@
...
+    await db.schema.createTable('principals', (table) => {
+      table.increments('id').primary();
+      table.string('external_id').notNullable();
+      table.string('nickname').notNullable();
+      table.integer('type').notNullable();
+      table.boolean('active').notNullable();
+      table.timestamp('created_at').defaultTo(db.fn.now());
+      table.timestamp('modified_at').defaultTo(db.fn.now());
+      table.boolean('system').notNullable();
+    });
```
evert, line 24: Could we run all migrations on the test database when it starts? This ensures that tests break if the schema changes.

test/pagination/services.ts
```diff
@@ -0,0 +1,85 @@
...
+    for(let i = 1; i < 251; i++){
+      const data = {
+        external_id: i.toString(),
+        nickname: `User ${i}`,
+        type: 1,
+        active: 1,
+        created_at: Date.now(),
+        modified_at: Date.now(),
+        system: 0
+      };
```
evert, line 36: Also would it be possible to create these via `principalService.save()`

I'm a big fan of integration-style tests because you get more bang for your buck. More code gets tested, and the data that gets stored is more likely to be accurate with real usage.

evert (commented):
Thank you! I feel inspired. I've been putting off database tests and looking at this I feel silly for not diving in.

Makefile
Zen-cronic: implemented suggestion: `DB_DRIVER=sqlite DB_FILENAME=":memory:" npx tsx --test ${TEST_FILES}`

src/principal/service.ts
Zen-cronic: i've implemented a generic `search`

src/principal/service.ts
```diff
@@ -69,6 +70,37 @@ export class PrincipalService {
...
+    const hasNextPage = (offset + pageSize) < total;
+
+    const result = await db('principals')
+      .where(filters)
+      .limit(pageSize)
+      .offset(offset);
+
+    const items: T[] = [];
+    for (const principal of result) {
+      items.push(recordToModel(principal) as T);
```
Zen-cronic, line 92: as the generic `search` is implemented, we'd also need a generic `recordToModel` later

evert (approved):
Great work!

PR #641, opened 2025-03-06
#624: adds UserInfo type and service functions

usrrname:
# Changes

- adds UserInfo type in `src/types`
- Adds `findUserInfoByUser(user: User)` function to `src/user/service.ts`, which returns a `UserInfo` type.
- creates starter service functions to userService: `findUserInfoById`, `recordToModel`, `updateUserInfo`
- call `updateUserInfo` on user controller when `PUT /user:id` called
- Add fields to GET `/user/:id` endpoint. 
- Add fields to json-schema for validation.
- Support new fields when doing PUT /user/:id

https://github.com/curveball/a12n-server/issues/624


src/user/controller/item.ts
```diff
@@ -52,12 +52,16 @@ class UserController extends Controller {
...
     const user = await principalService.findByExternalId(ctx.params.id);
-    user.active = !!ctx.request.body.active;
-    user.nickname = ctx.request.body.nickname;
+  
+    if (user.type === 'user') {
+      user.active = !!ctx.request.body.active;
+      user.nickname = ctx.request.body.nickname;
 
+      await userService.updateUserInfo(user, ctx.request.body.userInfo);
+    }
```
usrrname, line 61: RE:

> Check failure on line 60 in src/user/controller/item.ts
> 
> GitHub Actions
> / Build
> 
> Property 'userInfo' does not exist on type 'PrincipalEdit'.

I didn't want to go modifying more stuff if this wasn't the right place to do that

src/user/service.ts
```diff
@@ -117,3 +118,70 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+ * @param userInfo - new UserInfo object to update the UserInfo record with.
+ * @returns The updated UserInfo record.
+ * @throws BadRequest - If the UserInfo record is not updated.
+ */
+export async function updateUserInfo(user: User, userInfo: UserInfo): Promise<UserInfo> {
+  
+  const result = await db('user_info')
+    .where({principal_id: user.id})
+    .update(userInfo)
+    .returning('*')
```
usrrname, line 151: hmm maybe it's `returning()` as `.returning('*')` returns an array...

src/oidc/format/json.ts
```diff
@@ -1,31 +1,11 @@
-import { resolve } from 'url';
```
evert, line 1: This file should not have changed, and basically should be reverted.

This file is responsible specifically for returning an OpenID formatted `userinfo` object.

It should eventually use our new user information, but it's format comes directly from a standard and changing the format means this is broken.

The purpose of files in the `format/` directory are basically rendering JSON.

src/types.ts
```diff
@@ -211,3 +211,25 @@ export type AppClient = {
...
+  preferred_username?: string;
+  phone_number?: string;
+  phone_number_verified?: boolean;
+  locale?: string;
+  updated_at: number;
+  picture?: string;
+  address?: string;
+  birthdate?: string;
+  created_at: number;
+  modified_at: number;
```
evert, line 234: This belongs back in `src/oidc/format/json.ts`.

We do need our own UserInfo object, but it uses camelCase like all other types in this file, and should only have properties for the specific things that you added to the database schema.

usrrname: Ohhh I think I'm starting to see the difference between when _ and camelcase is used...

src/user/service.ts
```diff
@@ -117,3 +118,70 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+  middleName: string;
+  givenName: string;
+  familyName: string;
+  birthDate?: string | null;
+  address: string;
+  locale: string;
+  createdAt: Date;
+  modifiedAt: Date;
+  zoneInfo?: string | null;
+}
```
evert, line 171: This is actually the type that belongs in `src/types.js`

src/user/service.ts
```diff
@@ -117,3 +118,70 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+export async function findUserInfoById(user: User): Promise<UserInfo> {
+
+  const result = await db('user_info')
+      .select()
+      .where({principal_id: user.id})
+      .first();
+
+  if (!result) throw new NotFound(`UserInfo for user "${user.id}" not found.`);
+
+  return toUserInfo(user, result);
```
evert, line 136: Should use `recordToModel`

src/user/service.ts
```diff
@@ -117,3 +118,70 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+    .returning('*')
+    .first();
+
+  if (!result) throw new BadRequest(`UserInfo for user "${user.id}" was not updated.`);
+
+  return toUserInfo(user, result);
+}
+
+export type UserInfoModel = {
+  sub: string;
```
evert, line 160: `sub` should not exist. Did it accidentally make it in the database model?

usrrname: Nope... did not make it in there. 
Mental model was twisted as I believed this model had to include everything already required in our JSON response

usrrname: twas not on the migration script, but hmmmm
Screenshot 2025-03-05 at 11 28 50 PM

src/types.ts
```diff
@@ -211,3 +211,20 @@ export type AppClient = {
...
+  phoneNumberVerified?: boolean;
+  email?: string;
+  emailVerified?: boolean;
+  familyName?: string;
+  birthDate?: string;
+  address?: string;
+  locale?: string;
+  createdAt: Date;
+  modifiedAt: Date;
+  zoneInfo?: string;
```
evert, line 229: I think none of these should allow undefined, but all of them should allow `null`. This is more consistent with other parts in the codebase.

src/types.ts
```diff
@@ -211,3 +211,20 @@ export type AppClient = {
...
   requirePkce: boolean;
 };
+
+export type UserInfo = {
+  name: string;
+  middleName?: string;
+  givenName?: string;
+  phoneNumber?: string;
+  phoneNumberVerified?: boolean;
+  email?: string;
```
evert, line 221: This type should only have the information that was actually added in the database migration. Things like `phoneNumber` and `email` exists in other part of the datamodel (it's in PrincipalIdentity).

src/types.ts
```diff
@@ -211,3 +211,20 @@ export type AppClient = {
...
+  emailVerified?: boolean;
+  familyName?: string;
+  birthDate?: string;
+  address?: string;
```
evert, line 225: `address` should use the format from the spec in ticket #623

src/types.ts
usrrname: Oh yeahh `null`, a database thing!~

src/user/service.ts
```diff
@@ -117,3 +117,92 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+    .returning('*')
+    .first()
+
+  if (!result) throw new BadRequest(`UserInfo for user "${user.id}" was not updated.`);
+
+  return recordToModel(user, result);
+}
+
+type UserInfoRecord = {
+  sub: string;
```
evert, line 159: This should use the type from ~/src/db-types.ts . See #642. I've taken the liberty to merge this into this branch already!

src/user/service.ts
```diff
@@ -117,3 +117,92 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+ * @param userInfo - new UserInfo object to update the UserInfo record with.
+ * @returns The updated UserInfo record.
+ * @throws BadRequest - If the UserInfo record is not updated.
+ */
+export async function updateUserInfo(user: User, userInfo: UserInfo): Promise<UserInfo> {
+  
+  const result = await db('user_info')
+    .where({principal_id: user.id})
+    .update(userInfo)
+    .returning('*')
```
evert, line 150: This unfortunately is a Postgres specific feature.

However, I don't think you have to return anything here. Most `update` functions don't, and the `put` request doesn't return its representation. So returning `<void>` is fine

src/user/service.ts
```diff
@@ -117,3 +117,92 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+  phone_number_verified?: boolean;
+  locale?: string;
+  updated_at: number;
+  picture?: string;
+  address?: string;
+  birthdate?: string;
+}
+  
+export async function recordToModel(user: User, record: UserInfoRecord): Promise<UserInfo> {  
+  const identities = await db('principal_identities')
```
evert, line 178: Incorrect database table

src/user/service.ts
```diff
@@ -117,3 +117,92 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+
+  for (const identity of identities) {
+    if (!record.email && identity.uri.startsWith('mailto:') && identity.is_primary) {
+      record.email = identity.uri.replace('mailto:', '');
+      record.email_verified = identity.verified_at ? true : false;
+    }
+    if (!record.phone_number && identity.uri.startsWith('tel:') && identity.is_primary) {
+      record.phone_number = identity.uri.replace('tel:', '');
+      record.phone_number_verified = identity.verified_at ? true : false;
+    }
```
evert, line 190: None of these fields exist in the database

src/user/service.ts
usrrname: woops... my rebase overwrote

src/types.ts
usrrname: good callout! removed!

src/types.ts
usrrname: Did you mean #624 ?

evert: Yes! oops

src/user/formats/hal.ts
```diff
@@ -83,7 +83,10 @@ export function item(user: User, privileges: PrivilegeMap, hasControl: boolean,
         href: group.href,
         title: group.nickname,
       })),
-
+      'userinfo': {
+        href: `${user.href}/userinfo`,
+        title: 'User Info'
+      },
```
evert, line 89: It's not needed to make a new endpoint / route for userinfo, because we're just going to return the data as part of the User resource.

src/user/formats/hal.ts
```diff
@@ -94,7 +97,8 @@ export function item(user: User, privileges: PrivilegeMap, hasControl: boolean,
     createdAt: user.createdAt.toISOString(),
     modifiedAt: user.modifiedAt.toISOString(),
     type: user.type,
-    privileges
+    privileges,
+    userInfo,
```
evert, line 101: Lets just make this part of the user response body instead of embedded in an object, so:

```suggestion
    ...userInfo,
```

src/user/formats/hal.ts
```diff
@@ -280,6 +284,9 @@ export function newUserResult(user: User, password: string|null, identities: Pri
       'me': identities.map( identity => (
         { href: identity.uri, title: user.nickname ?? undefined }
       )),
+      'userinfo': {
+        href: `${user.href}/userinfo`,
+      },
```
evert, line 289: Also not needed. This implies a brand new route / endpoint but the new data simply exists on the user resource.

src/user/service.ts
```diff
@@ -117,3 +118,65 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+      {
+        name: userInfo.name,
+        locale: userInfo.locale,
+        given_name: userInfo.givenName,
+        family_name: userInfo.familyName,
+        birthdate: userInfo.birthDate,
+        address: userInfo.address ? JSON.stringify(userInfo.address) : null,
+        zoneinfo: userInfo.zoneInfo,
+      }
+    );
```
evert, line 162: The issue with this function is that it assumes that a record already exists for a user, but this should also handle the case where there was no record yet for a user.

So to do this correctly, this should use `.insert()` instead of `.update()`, and then use `onConflict()` to to turn the `INSERT` into an `UPDATE` automatically

src/user/service.ts
```diff
@@ -117,3 +118,65 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+    .where({principal_id: user.id})
+    .update(
+      {
+        name: userInfo.name,
+        locale: userInfo.locale,
+        given_name: userInfo.givenName,
+        family_name: userInfo.familyName,
+        birthdate: userInfo.birthDate,
+        address: userInfo.address ? JSON.stringify(userInfo.address) : null,
+        zoneinfo: userInfo.zoneInfo,
```
evert, line 160: This should also set `updatedAt` to `Date.now()`.

src/user/service.ts
```diff
@@ -117,3 +118,65 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+  if (!result) throw new BadRequest(`UserInfo for user "${user.id}" was not updated.`);
+
+}
+
+
+export async function recordToModel(user: User, record: UserInfoRecord): Promise<UserInfo> {
+
+  return {
+    createdAt: record.created_at ? new Date(+record.created_at) : null,
+    modifiedAt: record.modified_at ? new Date(+record.modified_at) : null,
```
evert, line 173: Consider not returning these 2 properties. They're not super needed and you're going to run into a subtle issue later

src/user/controller/item.ts
evert: So there's a bit of a tricky thing here. Right now the `put()` action on resources supports both HTML forms and PUT requests using `appllication/json`, but with this more complex structure i think we need a different schema for each of these cases.

It's needed because if we don't, we don't have strict validation for any of this and that feels like a bad idea.

So i think I will spend some time tonight fixing this and I'll commit this directly to this branch.

In the end we'll have 2 JSON schemas, one for JSON and one for forms. So stay tuned

src/user/formats/hal.ts
usrrname: yeah I noticed actually this endpoint doesnt exist... but it's in fact at `http://localhost:8531/userinfo`?

evert: That endpoint is a OICD-specific endpoint and returns information about the currently logged in user. The format is also slightly different.

src/user/formats/hal.ts
usrrname: This `userInfo` also includes a `createdAt` and `modifiedAt`, which a spread operator would overwite on the user record... is that ok?

evert: Yeah this is why later in this review I suggested to just remove them. It's not needed for the client to have several date fields here, they're going to be the same usually anyway.

src/user/controller/item.ts
usrrname: you mean we need to have a JSON schema for `application/json`, and `multipart/form-data` right?

Yeah I agree with that

src/user/service.ts
usrrname: I did a global search and didn't see a precedence of `updatedAt` used in this codebase... is your intent is to have a non-namespace-collision use of `modifiedAt` on this `userInfo` type ?

src/user/formats/hal.ts
usrrname: yeah I noticed... wasn't clear to me whether this was an endpoint intended for returning web-standard-complicant userinfo  JSON to front end only for a logged in user or otherwise.

src/user/controller/item.ts
evert: Yeah exactly. They're so different now. We could get away with having 1 before

src/user/service.ts
evert: whoops I meant `modifiedAt` sorry!

src/user/service.ts
```diff
@@ -117,3 +118,84 @@ export async function validateUserCredentials(user: User, password: string, log:
...
+  };
+
+  const result = await db('user_info').where({principal_id: user.id}).update(data);
+
+  if (result === 0) {
+    // No rows for existing user_info was found, so insert a new record
+    await db('user_info')
+      .insert({
+        ...data,
+        modified_at: Date.now(),
```
evert, line 168: `created_at` should also be added to this one!

src/api-types.ts
```diff
@@ -56,27 +60,27 @@ export interface AppClientNewFormBody {
...
   /**
    * Is the client allowed to refresh tokens
    */
-  allowRefreshToken?: "on";
+  allowRefreshToken?: string;
   /**
    * Can the client can use the 'password' flow flow
    */
-  allowPassword?: "on";
+  allowPassword?: string;
```
evert, line 83: Ignore all these. looks like `api-types` wasn't fully updated, so this automatically got updated too. But it's not relevant for your PR

usrrname: This was the same type changes it gave me when I ran them

usrrname: I see we have tech debt! !!!! 😱

evert: lol yes. plenty haha

src/user/controller/item.ts
```diff
@@ -52,12 +54,16 @@ class UserController extends Controller {
...
 
     const user = await principalService.findByExternalId(ctx.params.id);
-    user.active = !!ctx.request.body.active;
-    user.nickname = ctx.request.body.nickname;
+
+    if (user.type === 'user') {
+      user.active = !!ctx.request.body.active;
+      user.nickname = ctx.request.body.nickname;
+
+      await userService.updateUserInfo(user, ctx.request?.body?.userInfo as UserInfo);
```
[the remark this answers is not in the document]
evert: Yes! So you'll want to test this with postman, insomnia or whatever the tool du jour is =)

src/user/service.ts
```diff
@@ -127,26 +127,23 @@ export async function validateUserCredentials(user: User, password: string, log:
...
-      modified_at: Date.now(),
-    });
-    //@ts-expect-error - Create a partial UserInfoRecord for inserting any
-    // fields that could link a user to a userInfo record
-    result = {  
-      principal_id: id,
-      given_name: user.nickname,
-      created_at: Date.now(),
-      modified_at: Date.now(),
+    // No user_info record found, so return an empty UserInfo object
```
usrrname, line 136: OOHHHH

src/user/service.ts
```diff
@@ -159,70 +156,47 @@ export async function findUserInfoByUser(user: User): Promise<UserInfo> {
...
+    locale: record.locale,
+    givenName: record.given_name,
+    middleName: record.middle_name,
+    familyName: record.family_name,
+    birthdate: record.birthdate,
     address: record.address ? JSON.parse(record.address) : null,
-    zoneInfo: record.zoneinfo || null,
-    modifiedAt: record.modified_at ? new Date(+record.modified_at) : null,
+    zoneinfo: record.zoneinfo,
+    metadata: record.metadata ? JSON.parse(record.metadata) : {},
```
usrrname: At this point, I assume if none of the values exist, theyd already be `null` due to the database response so we dont have to fallback or `??`

evert: Yes, that's why I removed most of the `??` here. I kept the extra check for metadata just in case someone does something funky with that field and tries to clear it by setting it to an empty string or null. I guess I don't 100% trust that there's valid JSON in there but probably that's silly?

PR #642, opened 2025-03-06
Re-genrated db-types.ts file

evert:
Hey @usrrname . I ran the generate-db-types.mjs script, and re-built this file. (it can only be done with MySQL right now, so I thought it was easier to just run this).


usrrname (approved):
I think you have to lint?

usrrname:
weird, I was so curious why my types wouldnt update!

Issue #646, opened 2025-03-14
Add seed data for users


usrrname:
As a lazy and impatient developer, I would like to be able to use a seed function to have a pre-existing set of dummy users when I am developing with a12n-server

The seed data should be able to populate the following kinds of servers:
- mysql 
- sqllite
- pgsql

PR #647, opened 2025-03-14
#646 feat: add seed data for dummy users

usrrname:
# Changes

- adds `seeds/` folder and script for opting in to seed dummy users

- include `env_file` property on docker-compose to read from `.env` variable changes a user might make during development

Testing:

Screenshot 2025-03-24 at 9 55 59 AM

Screenshot 2025-03-24 at 9 55 07 AM

created `jane.smith@gmail.com`, was able to log in with autogenerated password

#646

TODO: 
- [x] make use of seed opt-in


evert:
Was this tested if you got an error?

seeds/001_users.ts
```diff
@@ -0,0 +1,98 @@
...
+export async function seed(knex: Knex): Promise<void> {
+  // First, delete existing entries to start fresh
+  await knex('user_info').del();
+  await knex('user_passwords').del();
+  await knex('principals').del();
+
+  // Insert seed users
+  const users = await knex('principals').insert([
+    {
+      identity: 'admin@example.com',
```
evert, line 13: This field should no longer be used. It's from before there was a `principal_identity` table which is now used instead.

seeds/001_users.ts
```diff
@@ -0,0 +1,98 @@
+import * as bcrypt from 'bcrypt';
+import { Knex } from 'knex';
+
+export async function seed(knex: Knex): Promise<void> {
+  // First, delete existing entries to start fresh
+  await knex('user_info').del();
+  await knex('user_passwords').del();
+  await knex('principals').del();
```
evert, line 8: This feels like a really bad idea. Is this how seed files normally work?

seeds/001_users.ts
usrrname: LOL shite. how does one stay up to date with most current db/schema state? do they just look at what the table headings are when they load the datagrip/tableplus?

seeds/001_users.ts
usrrname: no... this was claude bs. I unquestionably agreed with it after seeing delete statements at the beginning of local dev schema files.

usrrname: However... you mentioning that does make me recall that you intended to set up local dev database to persist test users... so this approach is counter to your intent...

evert: Everything should already persist by default. I think it's just your docker container right now that doesn't but there's no problem with the normal dev setup for this i think.

seeds/001_users.ts
evert: well, the 'internal api' is really the specification. So it's always more reliable to call the principalService and 'save()' functions instead of directly inserting into the database.

This has the advantage that you get full typing, and it also continuous to work with code changes.

seeds/001_users.ts
usrrname: I dunno If it's just me but I think the users I created have been persisting... but so far I cant validate/verify them without supporting the mail part I think

evert: You're not able to see those users when you navigate to http://localhost:8531/user

seeds/001_users.ts
evert: Also for additional context, I didn't drop that column right away because *just* in case someone runs into a bug with that migration, it's nice if the old column is still around. So it's a bit of a precaution to keep the old schema around and then later on remove it once it had some time to settle.

usrrname: Makes sense from a backward compatibility perspective! Lol when I worked corporate we announced deprecations like 3 months in advance

seeds/001_users.ts
usrrname: I see the users, I just can't login as them as the system will claim they haven't verified (existing docker setup with 2 other manually entered registrations)

evert: That makes sense! Because email verification needs to happen

usrrname: Yep, that's what I meant, we don't have a way to support that in dev env yet

usrrname, line 8:
```suggestion
```

usrrname:
> Was this tested if you got an error?

Yep... working through these... tons of ` null value in column "field" of relation "table-name" violates not-null constraint`

usrrname:
Last night I did get something going in, but it was still breaking 
Screenshot 2025-03-16 at 9 05 40 PM

evert:
> Last night I did get something going in, but it was still breaking

I noticed this was force-pushed after. Was this still an issue? Don't 100% understand how that could happen!

usrrname:
The force-push was a result of me making changes, squashing a commit and pushing it up :p

usrrname:
Rebased... and still err-ing: 

of note: `PostgreSQL Database directory appears to contain a database; Skipping initialization` 
Seems like if there was already some data on a mounted volume, the migrations wont run.
<details><summary>Full Trace</summary>  

```
Attaching to a12n-server-1, db-1
db-1           | 
db-1           | PostgreSQL Database directory appears to contain a database; Skipping initialization
db-1           | 
db-1           | 2025-03-17 23:20:34.425 UTC [1] LOG:  starting PostgreSQL 17.2 (Debian 17.2-1.pgdg120+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
db-1           | 2025-03-17 23:20:34.425 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db-1           | 2025-03-17 23:20:34.425 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db-1           | 2025-03-17 23:20:34.426 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db-1           | 2025-03-17 23:20:34.429 UTC [29] LOG:  database system was shut down at 2025-03-17 23:19:02 UTC
db-1           | 2025-03-17 23:20:34.434 UTC [1] LOG:  database system is ready to accept connections
a12n-server-1  | ⚾ @curveball/a12n-server 0.30.0
a12n-server-1  | PUBLIC_URI environment variable was not set, defaulting to http://localhost:8531/
a12n-server-1  | Running database migrations
a12n-server-1  | Running database seeds
db-1           | 2025-03-17 23:20:35.294 UTC [33] ERROR:  column "principal_id" of relation "user_info" does not exist at character 73
db-1           | 2025-03-17 23:20:35.294 UTC [33] STATEMENT:  insert into "user_info" ("family_name", "given_name", "locale", "name", "principal_id") values ($1, $2, $3, $4, $5)
a12n-server-1  | Migrations failed Error: Error while executing "/opt/app/dist/seeds/001_users.js" seed: insert into "user_info" ("family_name", "given_name", "locale", "name", "principal_id") values ($1, $2, $3, $4, $5) - column "principal_id" of relation "user_info" does not exist
a12n-server-1  |     at Seeder._waterfallBatch (/opt/app/node_modules/knex/lib/migrations/seed/Seeder.js:118:23)
a12n-server-1  | error: insert into "user_info" ("family_name", "given_name", "locale", "name", "principal_id") values ($1, $2, $3, $4, $5) - column "principal_id" of relation "user_info" does not exist
a12n-server-1  |     at Parser.parseErrorMessage (/opt/app/node_modules/pg-protocol/dist/parser.js:283:98)
a12n-server-1  |     at Parser.handlePacket (/opt/app/node_modules/pg-protocol/dist/parser.js:122:29)
a12n-server-1  |     at Parser.parse (/opt/app/node_modules/pg-protocol/dist/parser.js:35:38)
a12n-server-1  |     at Socket.<anonymous> (/opt/app/node_modules/pg-protocol/dist/index.js:11:42)
a12n-server-1  |     at Socket.emit (node:events:524:28)
a12n-server-1  |     at addChunk (node:internal/streams/readable:561:12)
a12n-server-1  |     at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
a12n-server-1  |     at Readable.push (node:internal/streams/readable:392:5)
a12n-server-1  |     at TCP.onStreamRead (node:internal/stream_base_commons:191:23) {
a12n-server-1  |   original: error: insert into "user_info" ("family_name", "given_name", "locale", "name", "principal_id") values ($1, $2, $3, $4, $5) - column "principal_id" of relation "user_info" does not exist
a12n-server-1  |       at Parser.parseErrorMessage (/opt/app/node_modules/pg-protocol/dist/parser.js:283:98)
a12n-server-1  |       at Parser.handlePacket (/opt/app/node_modules/pg-protocol/dist/parser.js:122:29)
a12n-server-1  |       at Parser.parse (/opt/app/node_modules/pg-protocol/dist/parser.js:35:38)
a12n-server-1  |       at Socket.<anonymous> (/opt/app/node_modules/pg-protocol/dist/index.js:11:42)
a12n-server-1  |       at Socket.emit (node:events:524:28)
a12n-server-1  |       at addChunk (node:internal/streams/readable:561:12)
a12n-server-1  |       at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
a12n-server-1  |       at Readable.push (node:internal/streams/readable:392:5)
a12n-server-1  |       at TCP.onStreamRead (node:internal/stream_base_commons:191:23) {
a12n-server-1  |     length: 134,
a12n-server-1  |     severity: 'ERROR',
a12n-server-1  |     code: '42703',
a12n-server-1  |     detail: undefined,
a12n-server-1  |     hint: undefined,
a12n-server-1  |     position: '73',
a12n-server-1  |     internalPosition: undefined,
a12n-server-1  |     internalQuery: undefined,
a12n-server-1  |     where: undefined,
a12n-server-1  |     schema: undefined,
a12n-server-1  |     table: undefined,
a12n-server-1  |     column: undefined,
a12n-server-1  |     dataType: undefined,
a12n-server-1  |     constraint: undefined,
a12n-server-1  |     file: 'parse_target.c',
a12n-server-1  |     line: '1065',
a12n-server-1  |     routine: 'checkInsertTargets'
a12n-server-1  |   }
a12n-server-1  | }
a12n-server-1 exited with code 1
Gracefully stopping... (press Ctrl+C again to force)
```

</details>

usrrname:
Now.. it seems I can set `seed.user` / `process.env.SEED_USER` or `cors.allowOrigin`/`process.env.CORS_ALLOW_ORIGIN` in `.env` independently. However, if I set both, I'll get this error when I try to login with the admin user 

Screenshot 2025-03-22 at 2 53 47 PM

.env.defaults
```diff
@@ -14,6 +14,18 @@ KEEP_ALIVE_TIMEOUT_MS=
 # functioning correctly.
 CURVEBALL_ORIGIN="http://localhost:8531"
 
+# Allow CORS from any origin by running the seed script 002_cors_allowed_origin.js
+# ie. CORS_ALLOW_ORIGIN=["http://localhost:8531", "http://localhost:3000"]
+CORS_ALLOW_ORIGIN=
+
```
evert, line 20: I was gonna leave a comment that this shouldn't use seed files and instead use the built-in facility for loading config from .env, but it looks like you partially did this?

This is not the type of thing that should ever exists in a seed file. Remember that seed files are contributor only and just meant for dev data.

.env.defaults
```diff
@@ -14,6 +14,18 @@ KEEP_ALIVE_TIMEOUT_MS=
...
+# NOTE: Seed variables should only be set in development mode. 
+# Seed scripts can only run if tables don't already contain values.
+# Seed scripts WILL CLEAR EXISTING DATA.
+
+# Run user seed scripts only
+# Seed scripts WILL CLEAR EXISTING principals and principal_identities tables.
+SEED_USERS=
```
evert, line 27: I mentioned this earlier,  I think the seed scripts should only ever run by calling an explicit `make` target, not as part of the server script. Adding seed-data is a one-time choice a contributor might make, not something that's an 'always on flag' or server-configuration feature, which is what environment files are a bit more intended for.

You already did this with `make knex-seed`, so I really think that's enough.

src/database.ts
```diff
@@ -5,13 +5,21 @@ import { fileURLToPath } from 'node:url';
...
+      if (process.env.NODE_ENV === 'development') {
+        console.info('Running database seeds');
+        return db.seed.run();
+      }
+    })
+    .catch((error) => {
+      console.error('Migrations failed', error);
+      process.exit(1);
+    });
 }
```
evert, line 23: This should all be reverted.

src/database.ts
```diff
@@ -187,6 +224,7 @@ export function getSettings(): Knex.Config {
       loadExtensions: ['.js'],
     },
     pool,
+    seeds: isDevelopment ? seedInfo : undefined,
```
evert, line 227: It's ok for seeds to always be in here, but they shouldn't always run. So this doesn't need to be conditional.

.env.defaults
evert: Seed files should effectively behave as a second set of migration files, ran in order and maintaining consistency over time.

usrrname: The addition of `CORS_ALLOW_ORIGIN` is meant to benefit both end users and contributors by allowing them to have a quick way to manipulate those values without having to directly run database code

.env.defaults
usrrname: Ah I thought the server script command was necessary in order for this to work.

But accordingly if I'm changing the seed config in settings, it should run when the `knex seed:run ` is called I hope

src/database.ts
usrrname: So far these are only supposed to run IF the env variables are set and on development

src/database.ts
usrrname: Ah I thought the server script command was necessary in order for this to work.

But accordingly if I'm changing the seed config in settings, it should run when the `knex seed:run ` is called I hope

.env.defaults
usrrname: Putting this here doesn't make it "always on" though... it's so they know about what server env variables are at their disposal to set

src/database.ts
```diff
@@ -5,13 +5,21 @@ import { fileURLToPath } from 'node:url';
...
+    .then(() => {
+      if (process.env.NODE_ENV === 'development') {
+        console.info('Running database seeds');
+        return db.seed.run();
+      }
+    })
+    .catch((error) => {
+      console.error('Migrations failed', error);
+      process.exit(1);
+    });
```
usrrname, line 22:
```suggestion
```

.env.defaults
evert: Yes, 100% agree. This is a useful environment variable. I mentioned I wanted to add this at our last meeting. But the problem is that you implemented it with seed files. Seed files are *not* used by end-users.

.env.defaults
evert: I meant that turning `SEED_USERS` off after the fact does not remove users. This makes it not a 'flag' but an action.

src/database.ts
evert: Seed files should not automatically run, so this conditional is not useful. Keeping `seeds` always set here should not have an effect whether we're in dev or production mode.

.env.defaults
usrrname: Oooohhh!

.env.defaults
usrrname: So what may be the alternative or better practice?

usrrname: >  built-in facility for loading config from .env, but it looks like you partially did this?

Like... `src/env.ts`?

usrrname: Gonna leave CORS_ALLOW_ORIGIN for a differnet PR

src/database.ts
evert: Yes it should. `knex seed:run` should run all the seed files unconditionally. I think seed files are very useful to get a whole bunch of test data, but there really should be no need for any configuration. This PR just needs to be simplified a bit

.env.defaults
```diff
@@ -14,6 +14,15 @@ KEEP_ALIVE_TIMEOUT_MS=
...
 
+# NOTE: Seed variables should only be set in development mode. 
+# Seed scripts can only run if tables don't already contain values.
+# Seed scripts MAY CLEAR EXISTING DATA.
+
+# Run user seed scripts only
+# Seed users script will not run if there are existing duplicate users in the database.
+# Note: if you set this, removing this flag will not remove existing users.
+# SEED_USERS=
+
```
evert, line 25: Should still be reverted!

src/database.ts
```diff
@@ -173,10 +180,26 @@ export function getSettings(): Knex.Config {
...
+  };
+
+  if (isSeedUsersEnabled) {
+    seedConfig.directory = seedDirectory + '/users';
+    seedConfig.specific = '001_users.js';
+  } else {
+    // no env variable set, no seeding needed
+    seedConfig = undefined;
+  }
+
```
evert, line 202:
```suggestion
  const seedDirectory = path.dirname(fileURLToPath(import.meta.url)) + '/seeds';

  /** Knex Seed API @link {https://knexjs.org/guide/migrations.html#seed-api} */
  const seedConfig: Knex.SeederConfig | undefined = {
    directory: seedDirectory,
    loadExtensions: ['.js'],
  };
```

src/database.ts
```diff
@@ -187,6 +210,7 @@ export function getSettings(): Knex.Config {
       loadExtensions: ['.js'],
     },
     pool,
+    seeds: seedConfig ?? {},
```
evert, line 213:
```suggestion
    seeds: seedConfig,
```

src/database.ts
```diff
@@ -67,8 +65,17 @@ export async function insertAndGetId<T extends Record<string, any>> (
...
 
+/**
+ * Checks if an environment variable exists and is not empty.
+ * @param key - The key of the environment variable to check.
+ * @returns `true` if the environment variable exists and is not empty, `false` otherwise.
+ */
+export function envVarExists(key: string): boolean {
+  const value = process.env[key];
+  return value !== undefined && value !== '';
+};
```
evert, line 76: This is not used

src/seeds/users/001_users.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const initialUsers = [
+  {
+    nickname: 'admin',
+    given_name: 'Admin',
+    family_name: 'User',
+    privilege: 'admin',
+    resource: '*',
+    scope: '*',
+    external_id: randomUUID().slice(0, 6),
```
evert, line 13: Any chance we could use the same algorithm for these as the actual server also uses? 

https://github.com/curveball/a12n-server/blob/main/src/crypto.ts#L29

src/seeds/users/001_users.ts
```diff
@@ -0,0 +1,82 @@
...
+import { Knex } from 'knex';
+
+const initialUsers = [
+  {
+    nickname: 'admin',
+    given_name: 'Admin',
+    family_name: 'User',
+    privilege: 'admin',
+    resource: '*',
+    scope: '*',
```
evert, line 12: I don't think these 2 are used.

.github/CONTRIBUTING.md
```diff
@@ -68,6 +68,8 @@ Everything prepended with `_` is either a hidden file or a build artifact.
 └── tsconfig.json
 ```
 
+Note: The `MAKEFILE` contains all commands for codebase operations, while the `package.json` scripts contains basic commands for startup and build.
```
evert, line 71:
```suggestion
Note: The `Makefile` contains all commands for codebase operations, while the `package.json` scripts contains basic commands for startup and build. 
```

src/seeds/users/001_users.ts
usrrname: I was copying what came in through the initial admin registration 😜

docs/server-settings.md
```diff
@@ -25,4 +27,5 @@ To store a string, it needs to be surrounded by double-quotes. For example, a le
 | `webauthn`                  | String  | `enabled` | Whether webauthn is "enabled", "disabled" or "required".'
 | `webauthn.serviceName`      | String  | `a12n-server` | The service name that should appear in Webauthn dialogs.
 | `webauthn.expectedOrigin`   | String  | `null` | The "origin" of this server. This must be set for webauthn to work
-| `webauthn.relyingPartyId`   | String  | `null` | The origin of the application performing the login.
\ No newline at end of file
+| `webauthn.relyingPartyId`   | String  | `null` | The origin of the application performing the login.
+|`seed.users`| boolean | `null` | Whether to populate local database with dummy users | `SEED_USERS`
```
evert, line 31: This one is still left over ! I appreciate the fixes for the other missing env variables though!

src/database.ts
```diff
@@ -67,8 +65,17 @@ export async function insertAndGetId<T extends Record<string, any>> (
 
 }
 
+/**
+ * Checks if an environment variable exists and is not empty.
+ * @param key - The key of the environment variable to check.
+ * @returns `true` if the environment variable exists and is not empty, `false` otherwise.
+ */
+export function envVarExists(key: string): boolean {
```
evert, line 73: Should this still be in there?

evert (approved):

evert:
Lgtm!

usrrname:
Does not look good to me, but I had to go cook dinner LOL
couple loose ends... I was part way through writing feedback on the UofT code too

usrrname:
it works on my machine!

usrrname:
LOL WHY SQLITE WHY https://github.com/curveball/a12n-server/actions/runs/14046723175/job/39329127972

evert:
I think the most likely problem with that error is that maybe it tries to use a different database. If you connect with sqlite to a non-existing file, it just creates the database with 0 tables on the spot.

So might be another relative path issue. :')

usrrname:
FINALLY 🟢 🥗 🚥 📗 🍏

usrrname:
This passed and all green, but unable to merge because: 

Screenshot 2025-05-10 at 5 13 20 PM

.github/workflows/test.yml
```diff
@@ -12,6 +12,9 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    strategy:
```
evert, line 15: Hey is there any reason you added all these?

This increases the number of github jobs for any PR to 24, which is quite large and can take a long time to complete. It's the most complete way to to test, but unittests themselves and the typescript build does already run on these 3 versions.

.github/workflows/test.yml
```diff
@@ -103,10 +124,20 @@ jobs:
         PG_DATABASE: a12nserver
         PG_HOST: 127.0.0.1
       run: make knex-migrate
+    - name: Test seed migrations
+      env:
```
evert, line 128: I noticed these environment variables are duplicated. Is there a way these can be specified in a way that they are re-used across steps? Maybe `env` can be specified on the job level?

evert (commented):
Very minimal last 2 comments, but mostly good to go! Was a long road but looks great

evert:
This should definitely be a squashed merge btw !

.github/workflows/test.yml
usrrname: I dont remember why I did...  perhaps it was a debugging step... I can remove them

usrrname: Dropped the commit

.github/workflows/test.yml
usrrname: I think they might be needed on every one of these, seems like I borked it by refactoring them onto the parent job

evert:
I think it's because the .yml is invalid :')

usrrname:
Ohhh yeah I remember why I added the extra `strategy` with node versions now!

Screenshot 2025-05-14 at 9 44 54 PM

.github/workflows/test.yml
```diff
@@ -6,7 +6,9 @@ on:
   push:
     branches: [ main ]
   pull_request:
- 
+permissions:
+  contents: read
+  checks: write
```
usrrname, line 11: Added due to seeing this CodeQL warning 

Screenshot 2025-05-19 at 8 44 22 PM

usrrname:
Screenshot 2025-05-21 at 11 22 37 AM
I think this one is still there because that is what's on the `test.yml` on `main`

evert (approved):
Lets go!

PR #685, opened 2025-08-06
fix: export missing types

usrrname:
The purpose of this PR is to export types from `src/types.ts`. 

Context:

I found that I wanted to import `ServerStats` and `UserInfo` from here in order to maintain a contract with the data being presented from API responses on `a12n-server-admin`, however I found I was unable to import these types.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,5 +5,6 @@
  */
 export { default as mainMw } from './main-mw.ts';
 export { load as init } from './server-settings.ts';
-
 import './db-types.ts';
+import './types.ts';
+


src/index.ts
```diff
@@ -5,5 +5,6 @@
  */
 export { default as mainMw } from './main-mw.ts';
 export { load as init } from './server-settings.ts';
-
 import './db-types.ts';
+export * from './types.ts';
```
evert, line 9: Ideally we only export `api-types.ts`. I know that doesn't solve your immediate problem, but only api-types.ts is the 'server contract' that we publish.

If you're open to it, you could make a small json-schema for 'ServerStats`, (you can re-generate api-types using a script in `bin/`)

`types.ts` is internal and if we export it, changing those types means we're breaking backwards compatibility.

usrrname: Oh LOL! I based on it named as `src/types.ts` I thought it was authoritative but just missing
yeah that sounds good

PR #687, opened 2025-08-07
fix: export contract types (`api-types.ts`) 

usrrname:
😬 Woops, was still unable to import types as they were not exported from `src/index.ts`  🐸 
Also added comments to clarify internal types for project

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,7 +7,8 @@
 export { default as mainMw } from './main-mw.ts';
 export { load as init } from './server-settings.ts';
 
-/** db-types and src/types are internal to this project */
+// db-types.ts registers its types globally and is not explicitly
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -3,7 +3,12 @@
  * exposes a middleware for curveball, and a load function that
  * needs to be called to initialize the server
  */
+
 export { default as mainMw } from './main-mw.ts';
 export { load as init } from './server-settings.ts';
 
+/** db-types and src/types are internal to this project */
```
evert, line 10:
```suggestion
// Ensuring that db-types.ts gets loaded. This file registers its types globally are
// not explicitly imported by other files. This ensures that Typescript doesn't miss these.
```

evert (approved):

